Issue #2, opened 2024-07-18
Auth - just a thought!


evert:
Hey if you will need to have a way to create users and authenticate them, I'm happy to integrate this for you if I can use my own server for this:

https://github.com/curveball/a12n-server

It's probably imperfect but would be a good use-case for me and I'll be there for the initial implementation and support. Also no worries if the answer is no!

kbventures:
@evert the only solution we are considering so far is https://github.com/curveball/a12n-server but we haven't gotten to that phase of the project yet.

PR #21, opened 2024-11-04
Feature: add initial draft for data schema

madcampos:
Here is an initial draft of a JSON schema for the data we will be saving on the VMS, it is intended to be readable and not very complex.

Please provide feedback and discussion points on it.


data/volunteer.schema.json
```diff
@@ -0,0 +1,112 @@
...
+		"pronouns": {
+			"type": "string",
+			"description": "The volunteer preffered pronouns, if they wish to provide it."
+		},
+		"socialMedia": {
+			"type": "object",
+			"description": "A list of social media links/handles for reaching out to the volunteer.",
+			"required": ["slack"],
+			"properties": {
+				"slack": { "type": "string", "format": "uri" },
```
evert, line 81: One idea is to call this 'links' instead of 'socialMedia', make it an array and give each item in the array the following properties:

* href
* title
* rel

Now it's a general links object, but you can still find out what 'social network' it was by looking at the domain for example. But it's completely unrestricted. `rel` should be `me` for most of those.

madcampos: That is a good structure! What about, if we only allow for `rel=me` that we have a flag `isRelMe` instead of an open attribute?

Are there other values for the `rel` attribute that would be interesting to have?

evert: I'm trying to sneak a hypermedia format in here :P But there's lots of others:

https://www.iana.org/assignments/link-relations/link-relations.xhtml

If you only want to support `me` I would probably recommend `"const": "me"`. If all of these are technically links to external profiles they are probably all `"me"`.

evert: But one could be "friend" from this list: https://microformats.org/wiki/existing-rel-values

madcampos: Technically, all would be links to that person...

Yeah, we can definitely make it more generic. I just think would be nice to have a strong default and offer extensions if needed.

dxnn: I like this idea 👍

dxnn:
Cool, this is a great start for a volunteer profile object. 

One could also image a team profile object. Then, instead of adding teams inside an array in the volunteer object, we can have a separate type of JSON object that works like a join table:
`{ teamid, volunteerid, role, season, event_added }`

Then we can pull those events into both the team's page and the volunteer's page when we render them.

data/volunteer.schema.json
```diff
@@ -0,0 +1,112 @@
...
+		"birthday": {
+			"type": "string",
+			"format": "date",
+			"description": "The volunteer birthday, consisting of day and month, for celebrating \"birthdays of the month\" and other related events."
+		},
+		"joinedAt": {
+			"type": "string",
+			"format": "date-time",
+			"description": "The date the volunteer has joined Toronto JS."
+		},
```
dxnn, line 100: Maybe `joinedAt` isn't needed if we also create an event for them joining e.g. the "volunteer team"

dxnn:
It might also be worth having two different kinds of volunteer objects: one for public things, and another for private things, like email, birthday, and agreements. (I'm still trying to figure out how to do this with all the data being publicly readable, so we don't have to deal with auth...)

madcampos:
Addressing @dxnn's comments:
> One could also image a team profile object. Then, instead of adding teams inside an array in the volunteer object, we can have a separate type of JSON object that works like a join table:
{ teamid, volunteerid, role, season, event_added }
>
> Then we can pull those events into both the team's page and the volunteer's page when we render them.

We will definitely benefit from also having a `team` object. I think keeping them cross-referenced would be a good idea, like, keep a list of teams in the volunteer object and a list of volunteers in the team object. That way when searching for data it is faster than going through all objects.

---

> It might also be worth having two different kinds of volunteer objects: one for public things, and another for private things, like email, birthday, and agreements. (I'm still trying to figure out how to do this with all the data being publicly readable, so we don't have to deal with auth...)

We could do something like a JWT and have every private object be encrypted.

Another solution would be to have a SQL database and have a "build step", where when data is inserted, it also outputs and save a json file. The db would be the source of truth with a public mirror of some data in json format.

kbventures:
Privacy-sensitive field solutions may become fixed after building the MVP. There are built-in solutions for PostgreSQL and application-level ones, as mentioned earlier.

andrewmatte:
I think it should be teamMemberships which allows for multiple teams and sessions across teams instead of a one to one relationship

madcampos:
Hey folks, from today's meeting I made some changes to the schema (the meeting notes are on the file: `docs/Meeting Notes/2024-11-11.md`).

There was a simplification and generalization of some things for the MVP.

I've split the main schema into 3 smaller ones: `Profile`, `Team`, and `Role`.
They represent a volunteer's profile, a team within TJS and then a role that associates volunteers to teams. There are no restrictions on the numbers for this, so we have a many-to-many relationship.

URLs came into discussion, and we decided to opt for a more generic approach of allowing people to write a description for their profiles in markdown, they can then put whatever URLs and however they want.

The reasoning behind this was that we don't know yet what we will need in the future and to move fast with the MVP we'd rather skip this for now and add later when we have a better understanding of the needed data than start off with some data that we would have to maintain afterwards.

Another addition to have a log of all the things that happened is the `EventLog`. It provides a simple yet powerful way of keeping a record of everything going on like people joining teams, being assigned roles, leaving teams, etc.

There are two timestamps added to all schemas: `createdAt`, `insertedAt`. They represent the moment in time the thing has happened and the moment in time it was saved on the VMS. That gives us the ability of add data for things that happened in the past while still keeping track of when the data was added to the system.

The last field added to all schemas was `schemaVersion`. It is a simple integer to check if the data is in an "old" format and convert it. We think it will help in case we introduce breaking changes to the schemas.

evert:
FYI not sure if it's interesting but if people are still interested in #2 , a12n-server already has support for teams (groups / roles) and team-based permissions. Any other requirement you may have I'm happy to add. It's fun and a lot of the things discussed here make sense for my project too. I

madcampos:
To add more context, here is how each of the schemas should relate to each other:

Untitled Diagram drawio

madcampos:
Superseded by #49

Issue #64, opened 2024-12-16
Team CRUD


kbventures:
[#38](https://github.com/torontojs/vms/issues/63)

PR #68, opened 2024-12-26
add routes to handle create, delete and update team

heem42:
## Summary

Added routes to handle create, delete and update the teams

## Tests

Manually tested using Postman

## Additional information

Issue #64


src/controllers/sqlControllers/createTeam.ts
```diff
@@ -0,0 +1,39 @@
+import type { Context } from 'hono';
+import { v4 as uuidv4 } from 'uuid';
+
+import { SCHEMA_VERSION } from 'src/constants/db.ts';
+import { StatusCodes } from 'src/constants/status-codes.ts';
+import type { TeamCreateBody } from 'src/types/data/team';
+
+export const createTeamSql = async (context: Context<EnvironmentBindings>) => {
```
madcampos, line 8:
```suggestion
export function async createTeamSql(context: Context<EnvironmentBindings>) {
```
**nitpick:** ‏Being _very annoying_ here, please use function declarations instead of function expressions.

Although most of the time there are no meaningful difference between both, I tend to prefer function declarations over function expressions where possible, so I'm being very picky about this. 🙃

References on the differences:
- https://www.freecodecamp.org/news/function-declaration-vs-function-expression/
- https://stackoverflow.com/questions/1013385/what-is-the-difference-between-a-function-expression-vs-declaration-in-javascrip

src/controllers/sqlControllers/createTeam.ts
```diff
@@ -0,0 +1,39 @@
...
+	try {
+		const body: TeamCreateBody = await context.req.json();
+
+		if (!body.name) {
+			return context.json({ error: 'name is required to create a team' }, StatusCodes.INTERNAL_SERVER_ERROR);
```
madcampos, line 13: **suggestion:** On the same lines about separation of concerns, here would be better to throw an error instead of sending a response.

That way, the only part of the code that returns a response to the network is the request handler, every other function either returns or throws an error.

Also, on this case the status code should be a [`415 Unsupported Media Type`](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/415).

src/controllers/sqlControllers/createTeam.ts
```diff
@@ -0,0 +1,39 @@
...
+		}
+
+		if (!body.happenedAt) {
+			return context.json({ error: 'happenedAt is required to create a team' });
```
madcampos, line 17: **suggestion:** ‏Here again on the sense of "separation of concerns", what about we have all of those checks for validation in a `validateTeamRequestData` function?

Most probably we will want to run other validations, like, if the name is a string and is not empty, etc. So if we move those checks to a function, they will be all in one place and will be easier to manage.

src/controllers/sqlControllers/createTeam.ts
```diff
@@ -0,0 +1,39 @@
...
+		}
+
+		const id = uuidv4();
+		const insertedAt = new Date().toISOString();
+		const happenedAt = new Date(body.happenedAt).toISOString();
+
+		await context.env.database.prepare(
```
madcampos, line 24: **issue:** ‏Please add a variable that receives the [return value of the `run` method](https://developers.cloudflare.com/d1/worker-api/prepared-statements/#run). 

We need to check here if the insertion was successful, if it was not, we want to throw and error.

src/controllers/sqlControllers/createTeam.ts
```diff
@@ -0,0 +1,39 @@
...
+			'INSERT INTO team (id, name, schemaVersion, description, happenedAt, insertedAt) VALUES (?,?,?,?,?,?)'
+		)
+			.bind(id, body.name, SCHEMA_VERSION, body.description ?? '', happenedAt, insertedAt)
+			.run();
+		const createdTeam = await context.env.database.prepare(
+			'SELECT * FROM team WHERE id = ?'
+		)
+			.bind(id)
+			.first();
```
madcampos, line 33: **issue:** ‏This method call is not needed.

In this function we already have all the data to construct the JSON response, so it makes more sense to actually return the constructed JSON instead of re-requesting it from the database.

src/controllers/sqlControllers/deleteTeam.ts
```diff
@@ -0,0 +1,24 @@
+import type { Context } from 'hono';
+
+import { StatusCodes } from 'src/constants/status-codes.ts';
+
+export const deleteTeamById = async (
```
madcampos, line 5: **suggestion:** ‏Same suggestion as for team creation: we should split this into request handler, db handler, and validator.

This has to do with future things, most probably the logic for deleting a team will include some dependencies, like cleanup before deleting the team.

src/controllers/sqlControllers/getProfiles.ts
```diff
@@ -1,6 +1,6 @@
 import type { Context } from 'hono';
 import { StatusCodes } from '../../constants/status-codes.ts';
-import type { Profile } from '../../db/profile';
+import type { Profile } from 'src/types/data/profile';
```
madcampos, line 3: **nitpick:** ‏Please keep relative imports

src/controllers/sqlControllers/getTeams.ts
```diff
@@ -3,28 +3,26 @@ import { StatusCodes } from '../../constants/status-codes.ts';
 
 export const getAllTeamsSql = async (context: Context<EnvironmentBindings>) => {
```
madcampos, line 4: **suggestion:** ‏Again here the same comment: let's split the code into request handler, sql handler, and validator.

src/types/data/team.d.ts
```diff
@@ -0,0 +1,20 @@
+export interface Team {
+	id: string;
+	schemaVersion: number;
+	name: string;
+	description?: string;
+	happenedAt: string;
+	insertedAt: string;
+}
+
+export interface TeamCreateBody {
```
madcampos, line 10: **suggestion:** ‏Here is a good place to be clever and use Typescript's [`Pick` utility type](https://www.typescriptlang.org/docs/handbook/utility-types.html#picktype-keys). This will make both types be in sync, so if we change something on the `Team` type it will reflect on `TeamCreateBody`.

src/types/data/team.d.ts
```diff
@@ -0,0 +1,20 @@
...
+	name: string;
+	happenedAt: string;
+	description?: string;
+}
+
+export interface TeamUpdateBody {
```
madcampos, line 16: **suggestion:** ‏Same with `TeamCreateBody`.

src/controllers/sqlControllers/updateTeam.ts
```diff
@@ -0,0 +1,46 @@
+import type { Context } from 'hono';
+
+import { StatusCodes } from 'src/constants/status-codes.ts';
+import type { TeamUpdateBody } from 'src/types/data/team';
+
+export const updateTeamById = async (context: Context<EnvironmentBindings>) => {
```
madcampos, line 6: **suggestion:** ‏Same as before, specially here as we need strong validation. Split the function into request handler, sql handler, and validator.

madcampos:
@kbventures what are your thoughts on adding a validation lidrary, like [Zod](https://github.com/colinhacks/zod) or [ Superstruct](https://github.com/ianstormtaylor/superstruct)?

src/controllers/sqlControllers/createTeam.ts
kbventures: This information will need to eventually be found in our contribution guide.

src/controllers/sqlControllers/createTeam.ts
kbventures: Created issue to address this [#70](https://github.com/torontojs/vms/issues/70)

src/controllers/sqlControllers/getProfiles.ts
pmjuu: if you're using vscode, you can add `settings.json` like this below.
```
{
    "typescript.preferences.importModuleSpecifier": "relative"
}
```

src/controllers/sqlControllers/updateTeam.ts
kbventures: File split in 3 functions which will moved to individual files in future pull requests.

kbventures (commented):
Great work @heem42 

Normally we would wait to let you make the changes but this issue is blocking other road map items.

src/types/data/team.d.ts
madcampos, line 1: **nitpick:** Use tabs for formatting‏

heem42:
Thank you, @madcampos, for your thoughtful comments and to @kbventures for implementing those fixes. Apologies for the delay in responding—I've been tied up with other work. I really appreciate your efforts on this! Please let me know if there are any further changes required.

pmjuu (approved):

src/controllers/sqlControllers/profileSqlControllers/getProfiles.ts
```diff
@@ -0,0 +1,59 @@
+import { z } from 'zod';
```
sounmind, line 1: Have you installed zod? I can't find zod dependency added in the package.json in your PR

kbventures (approved):

PR #72, opened 2025-01-14
Feat: profile crud - create, update

### LINKED ISSUE #63: Profile CRUD
[#38](https://github.com/torontojs/vms/issues/38)

lylaminju:
## Summary

- [x] `createProfile`
- [x] `updateProfile`
- [x] check validation

## Tests

succeeded creating and updating a profile through Bruno

## Additional information

- ~~I wonder [handling result.error](https://github.com/torontojs/vms/compare/setup-d1-database...profile-crud?expand=1#diff-80a7a49288c73e1b3140c35ba71e46c2a9075b369964c36a335a180dd25c6dc4R32-R34) is reasonable way to handle (rare) error cases. I think most of the INSERT query errors will be thrown by `run()` method before the if statement is executed.~~


src/controllers/createProfile.ts
```diff
@@ -0,0 +1,40 @@
+import type { Context } from 'hono';
+import { StatusCodes } from 'src/constants/status-codes';
+import { SCHEMA_VERSION } from 'src/db/constants';
+import type { Profile } from 'src/types/data/profile';
+
+type CreateProfileRequestBody = Omit<Profile, 'id' | 'schemaVersion'>;
+
+export const createProfile = async (context: Context<EnvironmentBindings>) => {
```
madcampos, line 8: **nitpick:** ‏Please use a function declaration instead of an arrow function.

src/controllers/createProfile.ts
```diff
@@ -0,0 +1,40 @@
...
+		const insertedAt = new Date().toISOString();
+
+		const result = await context.env.database
+			.prepare(`
+				INSERT INTO profile (id, email, name, description, links, happenedAt, insertedAt, schemaVersion)
+				VALUES (?,?,?,?,?,?,?,?)`)
+			.bind(id, email, name, description, links, happenedAt, insertedAt, SCHEMA_VERSION)
+			.run();
+
+		if (result.error) {
```
madcampos, line 32: **praise:** ‏This is a good error handling!

src/controllers/createProfile.ts
```diff
@@ -0,0 +1,40 @@
...
+		const {
+			email,
+			name,
+			description,
+			links,
+			happenedAt
+		} = body;
+		const insertedAt = new Date().toISOString();
+
+		const result = await context.env.database
```
madcampos, line 25: **suggestion:** ‏Maybe here we could destructure the results...
```suggestion
		const { success, results } = await context.env.database
```

src/controllers/createProfile.ts
madcampos, line 32: Small correction based on the [docs](https://developers.cloudflare.com/d1/worker-api/return-object/#d1result).

```suggestion
		if (!result.success) {
```

src/controllers/createProfile.ts
```diff
@@ -0,0 +1,40 @@
...
+				INSERT INTO profile (id, email, name, description, links, happenedAt, insertedAt, schemaVersion)
+				VALUES (?,?,?,?,?,?,?,?)`)
+			.bind(id, email, name, description, links, happenedAt, insertedAt, SCHEMA_VERSION)
+			.run();
+
+		if (result.error) {
+			throw Error(`INSERT error: ${result.error}`);
+		}
+
+		return context.json({});
```
madcampos, line 36: **suggestion:** ‏Here we should return some information, like the new id for the profile. Some APIs return a copy of the full data, including generated fields when creating a new resource.

src/controllers/createProfile.ts
pmjuu: @madcampos Thanks for the suggestion!
Btw, the `results` is always empty for INSERT ([ref](https://developers.cloudflare.com/d1/worker-api/prepared-statements/#guidance-1))
Is there any reason that you suggest to destructure `results` ?

src/controllers/createProfile.ts
kbventures: @pmjuu for messaging examples:

[team-crud branch](https://github.com/torontojs/vms/tree/team-crud)

src/controllers/createProfile.ts
```diff
@@ -0,0 +1,40 @@
+import type { Context } from 'hono';
+import { StatusCodes } from 'src/constants/status-codes';
+import { SCHEMA_VERSION } from 'src/db/constants';
+import type { Profile } from 'src/types/data/profile';
+
+type CreateProfileRequestBody = Omit<Profile, 'id' | 'schemaVersion'>;
```
kbventures, line 6: @madcampos @pmjuu 

This works but we're working differently on the team-crud by utilizing interfaces instead of using "Omit".  We should pick one or the other for the project?

```
export interface Team {
  id: string;
  schemaVersion: number;
  name: string;
  description?: string;
  happenedAt: string;
  insertedAt: string;
}

export interface NewTeamData {
  name: string;
  happenedAt: string;
  description?: string;
}

export interface UpdateTeamData {
  name?: string;
  happenedAt?: string;
  description?: string;
}

export type BaseTeamData = Pick<Team, 'description' | 'happenedAt' | 'name'>;
```

kbventures (commented):
Great work.

src/controllers/createProfile.ts
kbventures: Mixing Omit and separate interfaces  can confuse contributors. Standardizing one approach improves readability and maintenance.

pmjuu: @kbventures Good point!

At first, I preferred to reduce duplicate code and using `Omit` seemed to make maintenance easier.
It'd be better to use `Omit` if the base interface changes frequently and automatic updates to related types are needed. But I don't think our project will have frequent changes.

Using separate interfaces provides explicit control and clear definitions, which enhances readability and prevents unwanted side effects over time.

So, I think using separate interfaces will work better. What do you think? @sounmind

sounmind: Sure, both are having their good & bad points but I think code consistency is more important than deciding this. Thanks for comments @kbventures !

src/controllers/createProfile.ts
```diff
@@ -0,0 +1,55 @@
...
+			database: context.env.database
+		});
+
+		if (!success) {
+			throw new Error('Insertion failed');
+		}
+
+		return context.json(
+			{ message: 'Profile created successfully', createdId: meta.last_row_id },
+			StatusCodes.CREATED
```
madcampos, line 38: **praise:** Nice use of the HTTP status code‏

src/controllers/updateProfile.ts
```diff
@@ -0,0 +1,94 @@
...
+					.map((issue) => issue.message)
+					.join(', ')
+			},
+			StatusCodes.BAD_REQUEST
+		);
+	}
+
+	const totalFieldsToUpdate = Object.keys(parsedBody).length;
+	/**  Body only has happenedAt */
+	if (totalFieldsToUpdate <= 1) {
```
madcampos, line 42: **suggestion:** This should be split into two checks:
1. check if there is no fields to update
2. check if the only field is `happenedAt`

Better be explicit here than handle edge cases afterwards

src/controllers/updateProfile.ts
```diff
@@ -0,0 +1,94 @@
...
+		return context.json(
+			{
+				message: 'No fields to update'
+			},
+			StatusCodes.BAD_REQUEST
+		);
+	}
+
+	const profileId = context.req.param('id');
+	const isProfileIdValid: boolean = await validateProfileId({
```
madcampos, line 52: **suggestion:** No need to type the variable here if it is getting assigned from the return type of a function, the variable type will be inferred. If the function return type is nice it will translate nicely to the variable as well.‏

src/db/profile.ts
```diff
@@ -0,0 +1,67 @@
...
+interface UpdateProfileParams {
+	payload: UpdateProfilePayload;
+	database: D1Database;
+}
+
+export async function updateProfile({
+	payload: { id, name, description, links, happenedAt },
+	database
+}: UpdateProfileParams) {
+	const data = { name, description, links, happenedAt };
```
madcampos, line 55: **suggestion:** It is not a good idea to destructure an object to then build another object with the same properties.

If you want to pick only a couple properties and then keep the rest on another object, do like this:
```typescript
const { prop1, prop2, ...rest } = sourceObject.
```

On this case, it may also make more sense to send two parameters:
1. id
2. All of the properties to update

madcampos (commented):
Please take a look at this PR: #78

Specifically, the files on `/src/routes/team` and `/src/routes/profile`. If you would like, we can have a pair coding session to go over the structure

src/controllers/createProfile.ts
madcampos: Hummm not any specifics, no.
I think here then you can use only `success`, as it will tell if the operation was successful or not

madcampos:
I've updated the project packages and propagated it to the current active branches, if you want, we can jump on a call and solve the merge conflicts.

src/controllers/updateProfile.ts
lylaminju: @madcampos I thought if the only field is `happendAt`, there is no fields to update.

Could you explain more about the difference?

src/controllers/updateProfile.ts
```diff
@@ -0,0 +1,92 @@
...
+import type { ZodError } from 'zod';
+import { z } from 'zod';
+
+import { updateProfile } from '../routes/profile/data.ts';
+import { StatusCodes } from '../utils/responses.ts';
+import { validateProfileId } from '../validator/profile';
+
+const UpdateProfileSchema = z.object({
+	name: z
+		.string({ required_error: 'Name is required' })
```
madcampos: Name is optional here, so does the `required_error` property make sense here?

src/routes/profile/data.ts
```diff
@@ -1,25 +1,90 @@
+import { SCHEMA_VERSION } from '../../constants/db.ts';
 import type { Profile } from './validation.ts';
 
 export async function getProfileById(database: D1Database, profileId: string) {
 	const { results } = await database
-		.prepare('SELECT * FROM profiles WHERE id = ?')
+		.prepare('SELECT * FROM profile WHERE id = ?')
```
madcampos: **suggestion**: What if we set the table names as constants?

Something like:
```typescript
// File: constants/db.ts

export enum DatabaseTables = {
	PROFILE = 'profile',
	TEAM = 'team'
	// ...
}

// Other files...
database.prepare(`SELECT * FROM ${DatabaseTables.PROFILE}`);
```

src/routes/profile/data.ts
```diff
@@ -1,25 +1,90 @@
...
+		Profile,
+		'description' | 'email' | 'happenedAt' | 'links' | 'name'
+	>;
+	database: D1Database;
+}
+
+export async function insertProfile({
+	payload: { email, name, description, links, happenedAt },
+	database
+}: InsertProfileParams) {
```
madcampos: **praise**: That is an awesome destructuring!

src/routes/profile/data.ts
```diff
@@ -1,25 +1,90 @@
...
+      VALUES (?,?,?,?,?,?,?,?)
+      `
+		)
+		.bind(
+			id,
+			email,
+			name,
+			description ?? null,
+			links ?? null,
+			happenedAt,
```
madcampos: **suggestion**: I'm thinking here we want to set `happenedAt` to be the same as `insertedAt`.

Like, `insertedAt` is when the row is added to the database, and `happenedAt` is when the person actually joined Toronto JS. For all new users, those will be the same, it will only be different for people who joined before the VMS goes live. I think for those cases we can just edit it manually as it will be a limited set number of users.

src/routes/profile/data.ts
```diff
@@ -1,25 +1,90 @@
...
+	const entries = Object.entries(data).filter(
+		([, value]) => value !== undefined
+	);
+	const setClause = entries.map(([key]) => `${key} = ?`).join(', ');
+	const values = [...entries.map(([, value]) => value), id];
+	const query = `UPDATE profile SET ${setClause} WHERE id = ?`;
+
+	return database
+		.prepare(query)
+		.bind(...values)
```
madcampos:
```suggestion
		.bind(...entries.map(([, value]) => value), id)
```

The list of values can be destructured here directly, skipping one variable.

src/routes/profile/index.ts
```diff
@@ -54,3 +56,7 @@ profileRoutes.delete('/:id', async (context: Context<EnvironmentBindings>) => {
 		return context.json({ error: err.message }, StatusCodes.INTERNAL_SERVER_ERROR);
 	}
 });
+
+profileRoutes.post('/', createProfile);
+
+profileRoutes.patch('/:id', updateProfileById);
```
madcampos: **suggestion**: let's move the functions from controllers to this file, so they follow the same pattern as the other functions.

src/controllers/createProfile.ts
```diff
@@ -0,0 +1,55 @@
...
+			payload: parsedBody,
+			database: context.env.database
+		});
+
+		if (!success) {
+			throw new Error('Insertion failed');
+		}
+
+		return context.json(
+			{ message: 'Profile created successfully', createdId: meta.last_row_id },
```
madcampos: **question**: I think `last_row_id` here is not the actuall profile uuid, but the [`rowid`](https://www.sqlite.org/lang_createtable.html#rowid)... It is an internal number that SQLite uses to keep track of things.

One alternative would be for `insertProfile` to return if the operation is successful and the new profile uuid, like this:
```typescript
async function insertProfile({ payload, database }) {
	const id = crypto.randomUUID();
	
	// ...
	
	const { success } = await database.run(/* ... */);

	return { success, id };
}
```

src/controllers/createProfile.ts
madcampos: Resolving this conversation as it is addressed by using a type inferred from the zod validation.

src/controllers/updateProfile.ts
lylaminju: Oh, that was a mistake while copying and pasting 😅 Thanks!

src/routes/profile/data.ts
lylaminju: Good idea! It prevents typos

src/controllers/updateProfile.ts
madcampos: For some reason my reply got lost in the void... Here it is again:

So, the check we have right now only checks for the number of keys of the object, a user may provide `name` and `description` for example and it would pass the check.

On the other hand, if the user passes only `name` it will not pass the check.

The check than makes more sense if it is for no field, and then checks to validate if the fields are correct.

lylaminju: @madcampos 
Actually, I assumed `happendAt` field is a required field, which means the datetime of update. (please correct me if I'm wrong)
I intended to count the number of keys of the object (including `happendAt`), and I thought there would be only one number if `parsedBody` has only `happendAt`.

But, I think just counting the keys of the object is not a good logic in this situation.
What about using `every()` to check if `parsedBody` has only the `happendAt`?
```js
const onlyHasHappenedAt = Object.keys(parsedBody).every(key => key === 'happenedAt');

if (onlyHasHappenedAt) {
  return context.json(
    { message: 'No fields to update' },
    StatusCodes.BAD_REQUEST
  );
}
```

src/routes/profile/data.ts
lylaminju: Ok, I agree. Then the client doesn't have to send `happendAt` in the request body when creating a profile.

src/controllers/updateProfile.ts
madcampos: So, let's break down the reasoning here...

## `happenedAt` being required
Yes, assuming `happenedAt` is required is _technically_ correct. That is, it is correct that the field is required on the database level and we want to always update it on the code.

But... Timestamps like those, that update every time, should not be left to the user to update, they have to be something the backend controls, or else the user can put whatever there and make the system misbehave.

For example, if the user inputs a time in the past or in the future, it will make things wrong. The logic will still be correct, but the data will be wrong.

## Validation options

Now there are two possible situations for validation here, in general:
1. All keys are optional but we need at least one key to be present
2. Some keys are required

Both can be handled by zod, but here are some considerations:

For 2, zod will error on the required keys already, making them well... required, so we won't need to worry about that.

For 1 however, we would need to implement a check like `Object.keys().length > 0` requiring at least _one_ key to be present. That can be done either in the code after the zod check has ran, or as a custom check to the `z.object` function.

---

## Side note

Although calling `Array.every` is a good solution for checking if a key exists, it may not be performant for biger objects. On javascript the performance difference between going through 1 or 1k itens on an array is _almost negligible_.

But, one option would be to have a check like this:
```typescript
if (Object.keys(obj).length === 1 && obj.happenedAt != undefined) {
  // ...
}
```

This is a lighter weight check because:
- `Object.keys` is optimized in C
- `Array.length` is a property access
- `!= undefined` is one check for the key presence, so another property access
- `Array.every` necessarily goes over every item on the array, increasing the complexity of the code

Ideally, if `Object.keys` is implemented natively as `O(1)`, then doing property access is also `O(1)`, compared to `Array.every` that makes the complexity grow to `O(n)`.

madcampos: My suggestions here:
- Add [`.partial`](https://zod.dev/?id=partial) to the schema, so it can be used to mark everything as optional.
- Add [`.refine`](https://zod.dev/?id=refine) to the schema as well for checking if at least one property is present with `Object.keys().length > 0`.

madcampos: Actually, @pmjuu, what about we merge this PR and change that one validation later?

@kbventures thoughts?

PR #76, opened 2025-01-19
Feature: update contributing guide

### LINKED ISSUE #71: Update CONTRIBUTION.MD
Add these points:
- Favoring function declaration when possible
- Utilize relative imports
- Error handling(The only part of the code that returns a response to the network is the request handler, every other function either returns or throws an error.)

madcampos:
## Summary

- Add notes to the contribution guide based based on some comments on previous PRs.
- Move coding standard to it's own folder

## Additional information

Resolves #71


rjtbansal (approved):
Nice! Good job

docs/Coding Standard/1 - Strict Mode and Linting.md
```diff
@@ -0,0 +1,25 @@
...
+
+## Typescript is set to strict mode
+
+The objective is to reduce potential errors and force to write defensive code that handles edge cases and weirdness from JavaScript.
+
+## Linting
+
+The code will be formatted before every commit and checked for linting errors before pushing.
+
+If you want to manually do any of those things, you can run the following commands:
```
bcha92, line 11: Suggestion: Maybe we can use "concurrently" to create a new script that can run all the below scripts into one convenient alias? I'm just using `npm run check` written below as an example to bundle `lint`, `typecheck`, `lint:js`, and `format` at the same time. But perhaps this could be a discussion for another time.

```json
"scripts": {
  "check": "concurrently \"npm run lint\" && \"tsc --noEmit\" && \"eslint --fix\" && \"dprint fmt --staged\" "
}
```

bcha92:
Other than this one quip I put as a comment (not as a review), we can polish some of the wording in the docs a bit, but for now this is a great starting point and I think it's good to go!

docs/Coding Standard/1 - Strict Mode and Linting.md
madcampos: From those, eslint with `--fix` edits the files, and dprint also does edit the files.

I think we can parallelize the other parts, but those should run in sequence to avoid race conditions like the two processes try to edit the same file. 😅

bcha92:
base branch out of date, should we update or rebase it?

madcampos:
@bcha92
> base branch out of date, should we update or rebase it?

Up to you. The only restriction is for merging pull requests we do not squash them.

abdulqshabbir (approved):
lgtm ❤️

PR #82, opened 2025-01-23
Feat: improve status codes

madcampos:
## Summary

Add description for status codes based on [MDN docs](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status).

## Additional information

Basically it is a copy of the documentation from MDN, so we can have it as an easy to use reference.

### DIFF
--- a/src/utils/responses.ts
+++ b/src/utils/responses.ts
@@ -1,10 +1,330 @@
+/**
+ * List of HTTP status codes adapted from [MDN Docs](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status).
+ */
 export enum StatusCodes {
-	NOT_FOUND = 404,
-	OKAY = 200,
+	/**
+	 * This interim response indicates that the client should continue the request or ignore the response if the request is already finished.
+	 */
+	CONTINUE = 100,
+	/**
+	 * This code is sent in response to an `Upgrade` request header from the client and indicates the protocol the server is switching to.
+	 */
+	SWITCHING_PROTOCOLS = 101,
+	/**
+	 * This code was used in _WebDAV_ contexts to indicate that a request has been received by the server, but no status was available at the time of the response.
+	 *
+	 * @deprecated
+	 * @variation WebDav
+	 * @see {@link https://developer.mozilla.org/en-US/docs/Glossary/WebDAV | WebDav}
+	 */
+	PROCESSING = 102,
+	/**
+	 * This status code is primarily intended to be used with the `Link` header, letting the user agent start [preloading](https://developer.mozilla.org/en-US/docs/Web/HTML/Attributes/rel/preload) resources while the server prepares a response or [preconnect](https://developer.mozilla.org/en-US/docs/Web/HTML/Attributes/rel/preconnect) to an origin from which the page will need resources.
+	 */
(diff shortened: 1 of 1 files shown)


Zen-cronic (commented):
Overall, i like the fact that having documentation in the repo can provide better developer experience. And because we don't need to worry about bundle size (this is a relatively small file in a relatively small repo), we surely can integrate all the enums.

But imho, we don't need **all** of those codes (especially the deprecated ones), and i think the devs will have an easier time using the intellisense from `StatusCode.` if we have less properties. 
How about we only include the most common codes - 200, 400, 500, etc. - for now and add more as we need?

manvinderjit (commented):
I agree, having the most common status codes like 201, 401, 403, 404, 503, 504 makes sense instead of all. Having specific codes makes response validation on the front-end easier. We can use generic codes 200, 300, 400 for uncommon status codes.

Additionally, how do we want to handle empty requests from the backend, i.e., the request was successful but the backend returned empty data? With a 2xx response or a 4xx response?

evert:
> Additionally, how do we want to handle empty requests from the backend, i.e., the request was successful but the backend returned empty data? With a 2xx response or a 4xx response?

Successful request status codes should always start with a 2.

manvinderjit:
> > Additionally, how do we want to handle empty requests from the backend, i.e., the request was successful but the backend returned empty data? With a 2xx response or a 4xx response?
> 
> Successful request status codes should always start with a 2.

Makes sense. Because the status code represents the network req/res cycle, which was successful. Although I have seen conflicting opinions on this, but I agree with this approach.

evert:
> Although I have seen conflicting opinions on this

I've also seen other takes on this, but they are wrong lol. The people that wrote the HTTP standards are alive, they disagree with this alternative take

lylaminju (commented):
> /**
	 * The server refuses the attempt to brew coffee with a teapot.
	 */
	I_AM_A_TEAPOT = 418,

Haha, there's a joke

---
I think it's okay to keep this file. The file size is small, and it can be useful for reviewing status codes as part of learning.

madcampos:
I will remove things that are uncommon, like the WebDav sepcific stuff and some deprecated status codes.

But answering this question by @manvinderjit:
> Additionally, how do we want to handle empty requests from the backend, i.e., the request was successful but the backend returned empty data? With a 2xx response or a 4xx response?

This is a case for a `200` response, it literally means the request was processed okay by the server with no errors (either from input data coming from the client or the server code itself). But still no data is a valid response.

E.g.: A search for a Code Club event in December 2024, we didn't have one on that month, so an empty response is completely valid here.

A `404` in contrast is a response for something that _should_ be there, but isn't.

E.g.: Requesting the profile information for an id that doesn't match any user. The request itself has a correct format, but the _data_ is incorrect.

Very subtle difference, but I hope it helped 😅

madcampos:
@pmjuu the 418 status code is indeed a [joke](https://en.wikipedia.org/wiki/April_Fools%27_Day_Request_for_Comments)!

I left it there, because it would be fun to have a request implemented for that as an easter egg LOL

lylaminju (approved):

manvinderjit (approved):

Zen-cronic (approved):

PR #83, opened 2025-01-29
Feat: auth service - sign-in, sign-up

lylaminju:
### Sign up
- `/auth/sign-up`
- create a new profile
  - store hashed password in DB
  - `activatedAt` is null as default
- send an email for account confirmation
  - email contains an `activationLink` with a hashed token

### Confirm account
- `/auth/activate`
- identify user's email by token in the url
- update `activatedAt`

### Sign in
- `/auth/sign-in`
- check `activatedAt` is not null
- store a session token in  KV store and return the token

## Tests

- used Bruno to test apis

### Hashing
- Used the [Web Crypto API](https://developer.mozilla.org/en-US/docs/Web/API/Web_Crypto_API)'s [PBKDF2](https://cheatsheetseries.owasp.org/cheatsheets/Password_Storage_Cheat_Sheet.html#pbkdf2)(Password-Based Key Derivation Function 2)
- `bcrypt` and `Argon2` are not supported in Cloudflare Workers

### Email
- You need an `SENDGRID_API_KEY` in `.dev.vars` file to send an email in local env
- When you sent an account confirmation email in local env, the logo image of TorontoJS in the email will not appear
  <details>
  <summary>Gmail modifies the image src url</summary>

  1. Gmail proxies all images to improve security and performance.
  2. When you use localhost, Gmail tries to fetch the image but fails because it’s not on a public server.
  3. The result is a broken image (`ci3.googleusercontent.com/...#http://localhost`), meaning Gmail couldn’t access it.
  </details>

## Reference

[OWASP Password Storage Cheat Sheet](https://cheatsheetseries.owasp.org/cheatsheets/Password_Storage_Cheat_Sheet.html)

[SendGrid email integration quickstart](https://www.twilio.com/docs/sendgrid/for-developers/sending-email/quickstart-nodejs#create-and-store-a-sendgrid-api-key)


src/routes/auth/data.ts
```diff
@@ -0,0 +1,10 @@
+import type { SignInData } from './validate.ts';
+
+export async function authenticate(database: D1Database, body: SignInData) {
+	const { results } = await database
+		.prepare('SELECT id FROM profile WHERE email = ? AND password = ?')
+		.bind(body.email, body.password)
```
evert: consider hashing passwords with bcrypt or argon2!

kbventures:
@madcampos @pmjuu continuning conversation from:

migration/wrangler-development-script.sh
```diff
@@ -4,14 +4,15 @@
 wrangler d1 execute vms-local-db --local --command "DROP TABLE IF EXISTS event_log; DROP TABLE IF EXISTS profile; DROP TABLE IF EXISTS role; DROP TABLE IF EXISTS team;"
 
 # Create tables
-wrangler d1 execute vms-local-db --local --command "CREATE TABLE IF NOT EXISTS profile (id TEXT PRIMARY KEY NOT NULL UNIQUE COLLATE BINARY, email TEXT NOT NULL UNIQUE, schemaVersion INTEGER NOT NULL, name TEXT NOT NULL, description TEXT, happenedAt TEXT NOT NULL, insertedAt TEXT NOT NULL, links TEXT);"
+wrangler d1 execute vms-local-db --local --command "CREATE TABLE IF NOT EXISTS profile (id TEXT PRIMARY KEY NOT NULL UNIQUE COLLATE BINARY, email TEXT NOT NULL UNIQUE, password TEXT NOT NULL UNIQUE, schemaVersion INTEGER NOT NULL, name TEXT NOT NULL, description TEXT, happenedAt TEXT NOT NULL, insertedAt TEXT NOT NULL, links TEXT, activatedAt TEXT);"
```
evert: The password having a UNIQUE flag is a bit weird. In theory it should be ok since passwords have salt, but it stands out as odd. Why are all the datetime field `TEXT` ?

src/email-templates/confirm-email.ts
```diff
@@ -0,0 +1,244 @@
...
+          text-decoration: none;
+          -ms-interpolation-mode: bicubic;
+        }
+
+        p {
+          display: block;
+          margin: 13px 0;
+        }
+      </style>
+      <!--[if mso]>
```
evert, line 47: This email template seems _very excessive_. Have you considered something simpler. They're really not needed to look like this anymore!

src/routes/auth/data.ts
kbventures: We decided to use PBKDF2 as it's the only native solution suitable for hashing passwords stored in a database. Other alternatives would require Rust-based implementations, which we couldn't get working.

src/routes/auth/index.ts
```diff
@@ -0,0 +1,147 @@
...
+
+export const authRoutes = new Hono();
+
+authRoutes.post('/sign-up', async (context: Context<EnvironmentBindings>) => {
+	const body = await context.req.json();
+	let parsedBody: CreateProfileRequestBody;
+
+	try {
+		parsedBody = CreateProfileSchema.parse(body);
+	} catch (error) {
```
evert, line 19: Are you certain this is the only error that can be thrown here? It's usually better to catch a specific error and rethrow if it wasn't that error.

src/routes/auth/index.ts
```diff
@@ -0,0 +1,147 @@
...
+		return context.json<StatusResponse>({ message: error?.message ?? 'Invalid input' }, StatusCodes.BAD_REQUEST);
```
evert: A 422 error is slightly preferable over 400 I think.

src/routes/auth/index.ts
```diff
@@ -0,0 +1,147 @@
...
+		return context.json({ message: 'Duplicate email' }, StatusCodes.CONFLICT);
+	}
+
+	// Create Profile
+	try {
+		// Generate salt (16 bytes)
+		const salt = crypto.getRandomValues(new Uint8Array(16));
+
+		// Hash password with PBKDF2
+		const hashedPassword = await hashPasswordPBKDF2(parsedBody.password, salt);
```
evert: 1. Why did you end up with a custom solution for this over a format with the salt baked in?
2. Consider moving the salt and concatenation logic in the utility function, outside the controller so at least all your hashing logic is centralized in one place.

src/routes/auth/index.ts
```diff
@@ -0,0 +1,147 @@
...
+
+		// Convert salt to Base64 for storage
+		const saltBase64 = btoa(String.fromCharCode(...salt));
```
evert: This feels red-flaggy. Is `TextEncoder` an option? I wouldn't take my chances with 2 encoding steps that technically should not be compatible. It's possible this is safe but I would rather use a utility that's designed for this specific purpose.

src/routes/auth/index.ts
```diff
@@ -0,0 +1,147 @@
...
+
+		// Store the hashed password and salt in the database
+		parsedBody.password = `${saltBase64}:${hashedPassword}`;
+		await insertProfile({ payload: parsedBody, database: context.env.database });
+	} catch (error) {
+		console.error('Error creating profile', error);
+		return context.json<StatusResponse>({ message: 'Error creating profile' }, StatusCodes.INTERNAL_SERVER_ERROR);
+	}
```
evert: Curious why you're catching unspecific errors. Wouldn't the default behaviour in Hono to return a `500` ? Don't catch unless you have a reason to.

src/routes/auth/index.ts
```diff
@@ -0,0 +1,147 @@
...
+		sgMail.setApiKey(context.env.SENDGRID_API_KEY);
+
+		const salt = crypto.getRandomValues(new Uint8Array(16));
+		const token = await hashPasswordPBKDF2(parsedBody.email, salt);
+		const encodedToken = encodeURIComponent(token);
+		await context.env.ACTIVATION_TOKENS.put(
+			encodedToken,
+			parsedBody.email,
+			{ expirationTtl: 60 * 10 }
+		);
```
evert: Is this some kind of global storage you're comparing against later? If yes, why are your secrets hashed. Can't they be random strings?

src/routes/auth/index.ts
```diff
@@ -0,0 +1,147 @@
...
+
+		await activateProfile(context.env.database, email);
+
+		// Remove token after successful activation
+		await context.env.ACTIVATION_TOKENS.delete(token);
+
+		return context.json({ message: 'Account activated successfully' }, StatusCodes.OKAY);
+	} catch (error) {
+		return context.json<StatusResponse>({ message: 'Error activating an account' }, StatusCodes.INTERNAL_SERVER_ERROR);
+	}
```
evert: Uneeded catch block

src/routes/auth/index.ts
```diff
@@ -0,0 +1,147 @@
...
+
+		// Extract salt and hashed password from stored format
+		const [saltBase64, storedHashedPassword] = storedPassword.split(':');
+		if (saltBase64 === undefined) {
+			throw new Error('Salt not found');
+		}
+		const salt = new Uint8Array([...atob(saltBase64)].map((c) => c.charCodeAt(0)));
+
+		// Hash input password with the same salt
+		const inputHashedPassword = await hashPasswordPBKDF2(parsedBody.password, salt);
```
evert: Move this to a password utility function too. I'd basically copy the `bcrypt` signature and expose 2 methods: `hashPassword` and `validatePassword`.

src/routes/auth/index.ts
```diff
@@ -0,0 +1,147 @@
...
+		}
+
+		// Generate session token
+		const sessionToken = crypto.randomUUID();
+		const hoursAhead = 1;
+		const tokenExpiry = String(Date.now() + hoursAhead * 60 * 60 * 1000);
+		const expiryAndUserEmail = `${tokenExpiry} ${parsedBody.email}`;
+		await context.env.SESSION_TOKENS.put(sessionToken, expiryAndUserEmail);
+
+		context.header('Set-Cookie', `auth_token=${sessionToken}; HttpOnly; Secure; SameSite=Strict; Expires=${tokenExpiry}; Path=/; Domain=torontojs.com`);
```
evert: Probably wouldn't put a domain in if you want people to test locally

src/routes/auth/index.ts
```diff
@@ -0,0 +1,147 @@
...
+
+		return context.json<StatusResponse>({ message: 'Authorized successfully', data: sessionToken }, StatusCodes.CREATED);
+	} catch (err) {
+		if (err instanceof ZodError) {
+			return context.json<StatusResponse>({ message: err?.message ?? 'Invalid input' }, StatusCodes.BAD_REQUEST);
```
evert: 422 > 400

src/routes/auth/index.ts
```diff
@@ -0,0 +1,147 @@
...
+		}
+		return context.json<StatusResponse>({ message: err?.message ?? 'An error has occurred' }, StatusCodes.INTERNAL_SERVER_ERROR);
```
evert: Don't catch generic errors

src/routes/profile/data.ts
```diff
@@ -26,25 +26,29 @@ export async function deleteProfileById(database: D1Database, profileId: string)
...
+	payload: Pick<
+		Profile,
+		'description' | 'email' | 'links' | 'name' | 'password'
+	>;
 	database: D1Database;
 }
 
 export async function insertProfile({
-	payload: { email, name, description, links },
+	payload: { email, name, password, description, links },
```
evert: Unrelated to this specific line but I think this file now leaks password hashes because `SELECT *` was used. Generally a good idea to avoid `SELECT *`. Alternatively you could move the password in a separate table.

src/utils/hashPassword.ts
```diff
@@ -0,0 +1,29 @@
+export async function hashPasswordPBKDF2(password: string, salt: Uint8Array): Promise<string> {
```
evert: Where does this algorithm come from. Definitely falls for me in the 'don't roll your own crypto' bucket. Without a source or comments this would not pass a code review for me. I fully admit it's because I don't understand it fully.

evert (commented):
Hope it's helpful!

src/routes/auth/index.ts
madcampos: `TextEncoder` should be an option on cloudflare workers

src/routes/auth/index.ts
madcampos: I think this is my fault, I did follow this pattern before.

As for default error catching behavior in Hono: I'm not sure how it does handle errors and exceptions, the docs are lacking in this regard...

src/routes/auth/index.ts
madcampos: 422 represents validation errors closer than 400?

src/routes/profile/data.ts
madcampos: You mean having a table with only ids and passwords? That sounds like a good idea...

package.json
```diff
@@ -11,10 +11,14 @@
 		"typecheck": "tsc --noEmit",
 		"lint:js": "eslint --fix",
 		"lint": "npm run typecheck && npm run lint:js",
-		"db:setup": "src/db/wrangler-development-script.sh"
+		"db:setup": "migration/wrangler-development-script.sh",
+		"db:setup:linux-file-auth": "chmod +x migration/wrangler-development-script.sh && ./migration/wrangler-development-script.sh",
```
madcampos: Git preserves file properties, or at least it should do so.

package.json
```diff
@@ -11,10 +11,14 @@
...
+		"db:get:profiles": "wrangler d1 execute vms-local-db --local --command 'SELECT * FROM profile'"
```
madcampos: **suggestion:** I'd suggest writing a script or exposing dev only endpoints instead of ‏adding more and more scripts to `package.json`.

src/email-templates/confirm-email.ts
```diff
@@ -0,0 +1,244 @@
+import type { Context } from 'hono';
+
+export function generateEmailHtml(context: Context<EnvironmentBindings>, token: string) {
+	const activationUrl = `${context.env.BASE_URL}/auth/activate?token=${token}`;
+	const logoUrl = `${context.env.BASE_URL}/assets/torontojs-logo.png`;
```
madcampos: **suggestion:** If this function is not using anything other than `BASE_URL` from `context`, don't pass the full object to the function. Pass only what it needs.‏

src/routes/auth/data.ts
```diff
@@ -0,0 +1,36 @@
+import { DBTables } from '../../constants/db.ts';
+import type { SignInData } from './validate.ts';
+
+export async function getPassword(database: D1Database, body: SignInData): Promise<string> {
+	const { results } = await database
+		.prepare(`SELECT password FROM ${DBTables.PROFILE} WHERE email = ? AND activatedAt IS NOT NULL`)
```
madcampos: **nitpick:** When querying for a single entry on the database add `LIMIT 1`.‏

src/routes/auth/data.ts
```diff
@@ -0,0 +1,36 @@
...
+	if (!password) {
+		throw new Error('No active account found with the provided email');
+	}
+
+	return password;
+}
+
+export async function checkEmail(database: D1Database, email: string) {
+	const { results } = await database
+		.prepare(`SELECT email FROM ${DBTables.PROFILE} WHERE email = ?`)
```
madcampos: **nitpick:** When querying for a single entry on the database add `LIMIT 1`.

src/routes/auth/data.ts
```diff
@@ -0,0 +1,36 @@
+import { DBTables } from '../../constants/db.ts';
+import type { SignInData } from './validate.ts';
+
+export async function getPassword(database: D1Database, body: SignInData): Promise<string> {
+	const { results } = await database
+		.prepare(`SELECT password FROM ${DBTables.PROFILE} WHERE email = ? AND activatedAt IS NOT NULL`)
+		.bind(body.email)
+		.run();
+
+	const password = results[0]?.['password'] as string | undefined;
```
madcampos: **nitpick:** ‏If the `run` call is typed, then there is no need to add a type assertion.

src/routes/auth/index.ts
madcampos: Ditto @evert here: the password handling and salt generation should be contained in a function, where it is given the password and return it hashed and salted.

src/routes/auth/index.ts
evert: I think you should generally not try to catch every error, but instead build a middleware higher up the chain that catches these and returns `500` for anything unexpected + anything else you want to do with the error.

Having all these different spots that swallow the underlying problem and send a blank 500 just causes you to lose information about bugs. So I would remove all unspecific catch blocks, which as a plus makes a lot of the controller code simpler and reduces a nesting level.

src/routes/auth/index.ts
evert: Yes, 422 is for: "I could parse the body (as JSON) but the values sent to didn't match". 415 is if it's the wrong media type (not JSON). `400` is good if for example JSON itself was invalid, but it's also the error code if you can't find a better, more specific 4xx code that suits.

src/routes/auth/index.ts
madcampos: > Why did you end up with a custom solution for this over a format with the salt baked in?

I think I have an answer to this question: it is due to a limitation on the platform, for some reason the best suited password hashing algorithm supported by the WebCrypto API, which cloudflare supports and exposes to us, is PBKDF2. And that algorithm has the salt as a [separate argument](https://developer.mozilla.org/en-US/docs/Web/API/Pbkdf2Params#salt), it is not builtin.

src/routes/profile/data.ts
evert: yeah it helps to compartmentalize a bit! It's not the worst to have it in the `users` table but the problem is that objects are passed through end-to-end and with `SELECT *` which is a bit dangerous. Ideally you explicitly pass each property at each layer (request body -> controller -> service -> controller -> response body)

src/routes/auth/index.ts
madcampos: Ditto: just generating a random string should be enough

src/routes/auth/index.ts
```diff
@@ -0,0 +1,147 @@
...
+			parsedBody.email,
+			{ expirationTtl: 60 * 10 }
+		);
+
+		const emailHtmlTemplate = generateEmailHtml(context, encodedToken);
+		const msg = {
+			to: parsedBody.email,
+			from: context.env.SENDER_EMAIL,
+			subject: '[TorontoJS] Confirm your account',
+			html: emailHtmlTemplate
```
madcampos: **nitpick:** We should add a text only fallback to html emails.‏

src/routes/auth/index.ts
```diff
@@ -0,0 +1,147 @@
...
+
+		// Verify if the input password matches the stored password
+		if (inputHashedPassword !== storedHashedPassword) {
+			return context.json<StatusResponse>({ message: 'Unauthorized' }, StatusCodes.UNAUTHORIZED);
+		}
+
+		// Generate session token
+		const sessionToken = crypto.randomUUID();
+		const hoursAhead = 1;
+		const tokenExpiry = String(Date.now() + hoursAhead * 60 * 60 * 1000);
```
madcampos: **thought:** ‏I wish we had temporal support yet 😭

src/types/global.d.ts
```diff
@@ -1,10 +1,12 @@
...
 interface EnvironmentBindings {
 	Bindings: {
-		database: D1Database
+		database: D1Database,
+		SESSION_TOKENS: KVNamespace,
+		ACTIVATION_TOKENS: KVNamespace,
+		BASE_URL: string,
+		SENDER_EMAIL: string,
+		SENDGRID_API_KEY: string,
+		ASSETS: Fetcher
```
madcampos, line 9: **question:** Where does those values come from?

Also, for environment variables, what if they are unset, are they still strings and return an empty string or are they `undefined`?‏

src/utils/hashPassword.ts
```diff
@@ -0,0 +1,29 @@
...
+		['deriveBits']
+	);
+
+	// Derive key using PBKDF2-HMAC-SHA256
+	const derivedKey = await crypto.subtle.deriveBits(
+		{
+			name: 'PBKDF2',
+			salt,
+			iterations: 600000,
+			hash: 'SHA-256'
```
madcampos: **nitpick:** I prefer if we use `SHA-512`, it is just larger‏

src/utils/hashPassword.ts
```diff
@@ -0,0 +1,29 @@
...
+		},
+		key,
+		256
+	);
+
+	// Convert derived key to Base64
+	const derivedKeyBase64 = btoa(String.fromCharCode(...new Uint8Array(derivedKey)));
```
madcampos: If we are using the node compatibility layer, I would try doing this:
```suggestion
	const derivedKeyBase64 = new Uint8Array(derivedKey).toString('base64');
```

Check if this works

src/utils/responses.ts
```diff
@@ -189,4 +189,5 @@ export enum StatusCodes {
 
 export interface StatusResponse {
 	message: string;
+	data?: string;
```
madcampos: **nitpick:** if it is a status response, we don't return any data. Create another interface instead for data responses‏

src/validator/profile.ts
```diff
@@ -7,7 +9,7 @@ export async function validateProfileId({
 }): Promise<boolean> {
 	try {
 		const { results } = await database
-			.prepare('SELECT profile.id FROM profile WHERE id = ?')
+			.prepare(`SELECT id FROM ${DBTables.PROFILE} WHERE id = ?`)
```
madcampos: **nitpick:** When querying for a single entry on the database add `LIMIT 1`.

__tests__/routes/auth/index.test.ts
```diff
@@ -0,0 +1,87 @@
...
+    }
+    return null;
+  },
+};
+
+const kvSessionMock = {
+	put : ( sessionToken: string, expiryAndUserEmail: string)=>{
+		return null;
+	}
+}
```
madcampos: To avoid this kind of "black box" testing, it makes more sense to write unit tests and pass whatever external resource handler the function needs as a parameter. i.e. use dependency injection.

migration/wrangler-development-script.sh
kbventures: They should all be DATETIME since we want them to follow ISO Date format. I made the change.

src/routes/auth/index.ts
kbventures: Native bindings in Node.js allow libraries to access system-level functionality or perform performance-critical tasks by using compiled code, typically written in C or C++. Cloudflare doesn't but it does give you access to WebCrypto API.  PBKDF2 is the only algorithm provided  in WebCrypto API that is suitable for hashing password which will be stored in a database. 

https://github.com/WICG/proposals/issues/59?utm_source=chatgpt.com

It's summary it's not great but it seems to be the best choice for this purpose based on our research.

kbventures: Native bindings in Node.js allow libraries to access system-level functionality or perform performance-critical tasks by using compiled code, typically written in C or C++. Cloudflare doesn't but it does give you access to WebCrypto API. PBKDF2 is the only algorithm provided in WebCrypto API that is suitable for hashing password which will be stored in a database.

https://github.com/WICG/proposals/issues/59?utm_source=chatgpt.com

It's summary it's not great but it seems to be the best choice for this purpose based on our research.

We do intend to move some of the logic into it's own reusable component and a middle ware needs to be created.

We are currently discussing what 'authorization' looks like in terms of organizers/administrators ability to make changes/give access.

src/routes/auth/index.ts
madcampos: Got it. That makes sense

src/routes/auth/index.ts
madcampos: Yes, that does make sense to have a "catch all" middleware at the very end.

I was thinking of having it more localized per controller, but yeah. It is better the way you are suggesting.

On a side note: I tested hono and it just sends a simple 500 error with a plain text body of "internal server error" for uncaught exceptions.
This is good enough 🤷‍♂️

migration/wrangler-development-script.sh
lylaminju: Yes, Hash collisions are extremely unlikely but possible, and the UNIQUE constraint does nothing to improve password security. I removed it!

src/routes/auth/index.ts
madcampos: I think the `NODE_ENV` variable is set, we can use that to determine and only add a domain if it is "production"

evert: Unless you have a reason to, you can just not set a cookie domain. By default it will be the origin.

src/routes/auth/index.ts
lylaminju: You're right! `const body = await context.req.json();` I think this might also throw error. (`SyntaxError`)
I revised to catch only specific error and rethrow others.

src/routes/auth/index.ts
lylaminju: I agree. I modified to use `crypto.randomUUID()` to generate a random string token

src/email-templates/confirm-email.ts
lylaminju: Actually, I just copied and pasted the html file from the design team. 
I tried to analyze, but I'm not sure what to do. 
Maybe the conditional comments for Microsoft Outlook (e.g., <!--[if mso]>) might be excessive?
Could you explain more about which part is excessive and how it could be improved? @evert @madcampos

src/types/global.d.ts
lylaminju: `BASE_URL` and `SENDER_EMAIL` come from the `wrangler.toml` file.
For now, I set my personal email as a default email for the local test..

https://github.com/torontojs/vms/blob/58d34b800f41e3f658195415e48a4489205857ca/wrangler.toml#L11-L15

lylaminju: For local env, `SENDGRID_API_KEY` is located in `.dev.vars`, which is gitignored and only used in local env.

For production env, we need to login to the Cloudflare account and set the API key using terminal command `wrangler secret put SENDGRID_API_KEY`
(I didn't do this yet, but we should do this before production deployment)

Check this doc [Secrets - Cloudflare Workers](https://developers.cloudflare.com/workers/configuration/secrets/)

lylaminju: `ASSETS` is for the [static assets in Cloudflare](https://developers.cloudflare.com/workers/static-assets/)
I added this to fetch torontojs logo png file in the email template. (I'm not entirely sure if this is an optimal way to fetch static file)

I referred to this [static-assets/binding/#runtime-api-reference](https://developers.cloudflare.com/workers/static-assets/binding/#runtime-api-reference)

lylaminju: @madcampos If the environment variable is empty, it is `undefined`

(I just tested api call without sendgrid api key in `.dev.vars`, and I found `undefined` in console log)

src/routes/auth/index.ts
lylaminju: Do you mean it would be better to set the expiry hour longer than just one hour?

src/routes/auth/data.ts
lylaminju: Good. I typed `run()` as `run<Profile>()`

src/utils/hashPassword.ts
lylaminju: Yeah, I found we already added `compatibility_flags = ["nodejs_compat"]` in `wrangler.toml`, but wondering how to use `Buffer.from().toString('base64')` in refactored version..

src/email-templates/confirm-email.ts
kbventures: @pmjuu and I tested 

``` <html lang="en">
<head>
  <meta charset="UTF-8">
  <meta name="viewport" content="width=device-width, initial-scale=1.0">
  <title>Confirm Your Account</title>
  <style type="text/css">
    body, html {
      margin: 0;
      padding: 0;
      height: 100%;
      width: 100%;
    }
    table {
      width: 100%;
      height: 100%;
      border: 0;
      cell-spacing: 0;
      cell-padding: 0;
    }
    .email-container {
      width: 600px;
      margin: 0 auto;
      background-color: #FFFFFF;
      padding: 20px;
      text-align: center;
      border-radius: 8px;
    }
    h1 {
      font-size: 24px;
      color: #333333;
    }
    p {
      font-size: 16px;
      color: #555555;
    }
    .button {
      display: inline-block;
      background-color: #ED342F;
      color: white;
      padding: 12px 24px;
      text-align: center;
      text-decoration: none;
      border-radius: 5px;
      font-size: 16px;
      margin-top: 20px;
    }
    .footer {
      font-size: 12px;
      color: #777777;
      margin-top: 20px;
    }
  </style>
</head>
<body>
  <table role="presentation">
    <tr>
      <td align="center" valign="middle">
        <div class="email-container">
          <!-- Image added here -->
          <img alt="TorontoJS logo" height="auto" src="${logoUrl}" style="border:0;display:block;outline:none;text-decoration:none;height:auto;width:100%;font-size:13px;" width="128">
          <h1>Confirm Your Account</h1>
          <p>Hello! Click the button below to confirm your account.</p>
          <a href="${activationUrl}" class="button">Confirm My Email</a>
          <p class="footer">If you didn't request this, please ignore this email.</p>
        </div>
      </td>
    </tr>
  </table>
</body>
</html>
```

<strike>It works for google mail:</strike> The email below is sent with previous code

image (4)


But surprisingly Hotmail(outlook) BLOCKS it.  Which is incredible considering the daily volume of phishing scams and spam I get daily unhindered.

image (3)

kbventures (dismissed):

evert:
@kbventures Regarding the hotmail block, how did you test sending the email?

lylaminju:
@evert Actually, I sent the email. I ran the server in local env (with my personal SendGrid API key) and used Bruno to test sending emails

evert:
So it was sendgrid's STMP server? What was the `From:` address for the email that was blocked?

lylaminju:
@evert I'm using the [SendGrid API](https://www.twilio.com/docs/sendgrid/for-developers/sending-email/quickstart-nodejs#how-to-use-nodejs-to-send-an-api-email), not the SMTP

Although an SMTP server might be used internally as part of SendGrid’s infrastructure during the email delivery process, our code doesn’t directly connect to that server I guess.

That email `jsi04049@gmail.com` is just my personal email used for the test, which is a verified email in my SendGrid account.
We can set the email sender using only a verified email.

evert:
> @evert I'm using the [SendGrid API](https://www.twilio.com/docs/sendgrid/for-developers/sending-email/quickstart-nodejs#how-to-use-nodejs-to-send-an-api-email), not the SMTP

Makes sense! Sendgrid still uses SMTP in the end to deliver the email. 

> That email jsi04049@gmail.com is just my personal email used for the test, which is a verified email in my SendGrid account.

That's why Hotmail blocks it. It doesn't matter if Sendgrid verifies it. The `gmail.com` uses DNS to tell other SMTP servers (like Hotmails) who is allowed to send email on gmail's behalf. Sendgrid is not allowed to send email for `@gmail.com` addresses, so a block is expected.

Basically the owner of a domain (google in this case) can specify which servers are allowed to send emails with their address in `From`. If this weren't the case, it would be easier to impersonate others.

evert:
https://sendgrid.com/en-us/blog/dont-send-email-from-domains-you-dont-control

src/routes/auth/index.ts
madcampos: No, no, was just a comment that I wish we had support for the [Temporal API](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Temporal) it makes dealing with dates SOOOOO MUCH EASIER!

But unfortunately, it is not supported yet. At all. In any engine... 😭

src/types/global.d.ts
madcampos: Please put `SENDER_EMAIL` in `.dev.vars` as well, I know it is "public information", but I'd rather not use _your_ email for everything 🫠

madcampos: For setting the prod variables, don't worry about it. Either Ken or I can do it later.

I'll only ask you to add a new file, and commit it to github. The file will be `.dev.vars.example` and should contain all of the names of the variables that are set in `.dev.vars` but without any values. That way we have a reference file that anyone joining the project can copy and add their own secrets.

If possible, also add comments to the file documenting the variables there.

src/email-templates/confirm-email.ts
pmjuu: correction: this html code didn't work well 😅

src/utils/hashPassword.ts
madcampos: Just try it in a test endpoint and see if it breaks, can be a random test buffer 🤷‍♂️ from the string "test" or something

src/types/global.d.ts
lylaminju: Good idea! I put `SENDER_EMAIL` in `.dev.vars`, and added `.dev.vars.example` for the reference

__tests__/routes/auth/index.test.ts
lylaminju: @kbventures Why don’t we write test codes after this branch is merged? It feels like this PR is getting too big.

madcampos (dismissed):
Approving this PR with the following considerations:
- Review and document the password hashing functions.
- Test email sending now that the DNS entries are updated
- Review html template

madcampos (approved):

PR #85, opened 2025-02-06
Features: add openapi docs, improve bd schemas, and standardize responses

madcampos:
## Summary

Main features:

- Update the route definitions to generate openapi docs and to use zod schemas for validation.
- Update database schema to align more with designs
- Add some standardized APi responses
  - [HATEOAS](https://en.wikipedia.org/wiki/HATEOAS) responses using [HAL](https://en.wikipedia.org/wiki/Hypertext_Application_Language)
  - Status response consisting only of message and an optional error
  - Paginated response with fields needed to build pagination

## Tests

None 🫠

### Database schemas

- Add command to run migrations on the database using wrangler. (This creates all tables on a local db).
- Add sql migration scripts with all of the fields documented.
- Add sample seed script with very minimal data for testing. (This can definitely be improved!)
- Make schemas/types for data cascade in a way that helps to keep them in sync[^1].

### Standardized responses

- Add HATEOAS so the response can be used to link to other urls using a very simple schema. This basically means adding a `_links` object with some links to related things to the response.
- Add a status response that is used throughout for anything that represents a status message and does not contain any "data". The meaning of the response, like if it is an error or a success should be inferred by the http status code[^2].
- Add the skeleton for a pagination type of response, as well as `TODO` comments to actually implement pagination. (This PR is already doing a lot as it is... 🫠🙃)

### Openapi docs

- The format for writing endpoints has changed to use the `app.openapi` format instead of `app.<http method>`.
- `app.openapi` requires a call to `createRoute` to document the api route. This is strictly typed and have lots of type sorcery going on, so typescript will actually yell if things are not conforming to the types. That means some changes:
  - Using `res.json({} satisfies Type)` instead of `res.json<Type>()`. For some reasons beyond my feeble mind one works, the other don't. The end results are the same for us.
  - Changing `StatusCodes` to be a [`const const`](https://www.totaltypescript.com/concepts/as-const) instead of an `enum`. The usage is the same, but it behaves _slightly_ differently... [^3]
  - _EVERYTHING_ has to be a zod schema, both responses and request parameters. The good part is the openapi addition does validation for us; the bad part is that we can't write simple `type ...` or `interface ...` anymore, it _has to go through_ zod... 🫠

### Miscellaneous
- Shout out to @pmjuu by catching that we need to validate if an id exists on the db before updating or deleting it! I've made that validator a little more generic and applied it throughout the CRUD routes.
- I've added a check so a request the tries to create a profile with an existing email will error out.
- A couple eslint rules are relaxed so they are warns instead of errors or will not error out

[^1]: The schemas move as follows: main entity > create > update (E.g.: Team > Create Team > Update Team). So all properties exist for results when retrieving data, creation allows to add everything, and update makes most fields optional. There is an extra restriction to not change the id, schema version or insertion timestamps.

[^2]: Sending a 200 response with `{ isError: true }` is a sin and people who write this type of abomination should suffer by writing Java and XML servers for the rest of eternity! __*BE WARNED!*__

[^3]: While the `const` with the `as const` annotation is a plain js object that will suffer no conversion, an `enum` suffers some transformations by the typescript compiler. Some people prefer to avoid enums because they are both types (that can be removed) and actual code, so they are in a weird spot between ts and js. Here is a more in depth read about enums: https://2ality.com/2025/01/typescript-enum-patterns.html


rjtbansal:
In the PR template:
`Changing StatusCodes to be a [const const](https://www.totaltypescript.com/concepts/as-const) instead of an enum. The usage is the same, but it behaves slightly differently...` 

What is `const const`? Or did you just mean `const`?

migrations/0002_create-profile.sql
```diff
@@ -0,0 +1,67 @@
...
+	-- A text blurb the person can provide about themselves
+	description TEXT,
+	-- A flag indicating if the user is based on the Grater Toronto Area (GTA)
+	isBasedOnGTA INTEGER NOT NULL DEFAULT 1 CHECK(isBasedOnGTA IN (0, 1)),
+	-- A flag indicating if the user is available to join local/in-person events
+	canJoinLocalEvents INTEGER NOT NULL DEFAULT 1 CHECK(canJoinLocalEvents IN (0, 1)),
+	-- The pronouns the person identifies with
+	pronouns TEXT,
+	-- The person's birthday, saved as month and day only
+	-- In the format: YYYY-MM
```
rjtbansal: The format appears to be year and month instead of month and day as you suggest on line 26. Should this then be:
`MM-DD` ?

migrations/0002_create-profile.sql
```diff
@@ -0,0 +1,67 @@
...
+	birthday TEXT,
+	-- The user avatar reference,
+	avatar TEXT,
+	-- The avatar may be a url, or an id for a file hosted on the platform
+	avatarSource TEXT DEFAULT 'id' CHECK(avatarSource IS NULL OR avatarSource IN ('id', 'url')),
+	-- The date this person has joined Toronto JS
+	happenedAt TEXT NOT NULL,
```
rjtbansal: As someone FE focused more like a question for my understanding. 

We store dates as text and not as a particular date format on BE? Similar question where ever you are mentioning date but type is TEXT

src/routes/profile/data.ts
```diff
@@ -1,13 +1,49 @@
...
+import type { CreateProfileData, Profile, UpdateProfileData } from './validation.ts';
 
-export async function getProfileById(database: D1Database, profileId: string) {
+export async function insertProfile(database: D1Database, body: CreateProfileData) {
+	const id = crypto.randomUUID();
+	const insertedAt = new Date().toISOString();
+	const happenedAt = insertedAt;
+
+	const { success } = await database.prepare(`
+		INSERT INTO ${DBTables.PROFILE} (
```
rjtbansal, line 8: Curiosity: We are writing plain SQL here. What made us take this path (is it that our devs are comfortable with it, doing it for learning rather than ORMs like Knex?

rjtbansal (commented):
Great job! Had some questions and clarifications

migrations/0002_create-profile.sql
madcampos: Good catch! It was supposed to be month and day, I mistyped it.

madcampos:
>In the PR template:
> Changing StatusCodes to be a [const const](https://www.totaltypescript.com/concepts/as-const) instead of an enum. The usage is the same, but it behaves slightly differently...
>
> What is const const? Or did you just mean const?

There is a link to the typescript docs, in js a `const` is not entirely immutable, for example you can push items to an array declared using `const`.

In typescript however, there is the `as const` type assertion that marks that declaration as `readonly`, marking it in fact _immutable to typescript_.

That makes type inferences more reliable, as inferring the keys of an object yields the actual keys and not `string` as it would otherwise.

It also makes the object not extensible, so the data can't be changed and new keys can't be added.

So, the object would be "locked" but still usable like the enum definition, with key access and all.

migrations/0002_create-profile.sql
madcampos: This is a limitation to the [data types supported by SQLite](https://sqlite.org/datatype3.html#affinity_name_examples), it does understand if you set a column as `DATE`, but it will be converted to something else.

I'm thinking here, maybe we should use the specific data types...

src/routes/profile/data.ts
madcampos: So, we discussed using a query builder or ORM before...

The argument against ORMs is that in general they are "everything and the kitchen sink", so would lock us to using the ORM's specific way of writing things and bloating the project.

Then the option was using a query builder like Knex, and we were almost doing that, but Cloudflare D1, our db of choice for the project, is not supported by Knex. Other query builders exist but are not as popular.

So, we decided to just go with plain SQL prepared statements and use that as an opportunity for everyone to learn and improve their db skills.

src/routes/profile/validation.ts
```diff
@@ -1,39 +1,37 @@
...
+export const CreateProfileSchema = ProfileSchema.omit(BaseDBFieldsToOmit);
+
+export type CreateProfileData = z.infer<typeof CreateProfileSchema>;
+
+export const UpdateProfileSchema = CreateProfileSchema
+	.omit({ email: true })
+	.partial()
+	.refine(
+		(data) => Object.keys(data).length === 0,
+		{ message: 'At least one property is requires' }
```
lylaminju: typo: 'requires' -> 'required'

src/routes/profile/data.ts
```diff
@@ -1,13 +1,49 @@
...
+			id, schemaVersion,
+			happenedAt, insertedAt,
```
lylaminju: It looks great to utilize `BaseDBFieldsToOmit`! Then, how about replacing these four fields with the keys of `BaseDBFieldsToOmit` to ensure the consistency across all queries?

Maybe like `${Object.keys(BaseDBFieldsToOmit).join(', ')},` <sub>(don't forget the comma at the end!)</sub>

src/routes/profile/index.ts
```diff
@@ -1,156 +1,285 @@
...
+			const isProfileIdValid = await validateExistingId(context.env.database, DBTables.PROFILE, id);
 
-	if (onlyHasHappenedAt) {
-		return context.json<StatusResponse>(
-			{ message: 'No fields to update' },
-			StatusCodes.BAD_REQUEST
-		);
-	}
+			if (!isProfileIdValid) {
+				return context.json({ message: 'Invalid profile ID' } satisfies StatusResponse, StatusCodes.BAD_REQUEST);
```
lylaminju: suggestion: How about using `NOT_FOUND`? It indicates that the requested id is not found.

src/routes/profile/index.ts
```diff
@@ -1,156 +1,285 @@
...
+			const isDeleted = await deleteProfileById(context.env.database, id);
 
-	try {
-		const { success } = await updateProfile({
-			id: profileId,
-			data: parsedBody,
-			database: context.env.database
-		});
+			if (!isDeleted) {
+				return context.json({ message: 'Profile not deleted' } satisfies StatusResponse, StatusCodes.FORBIDDEN);
```
lylaminju: suggestion: How about using `INTERNAL_SERVER_ERROR`? 
Because [`FORBIDDEN`](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/403) usually means insufficient permissions to a resource or action

src/routes/team/validation.ts
```diff
@@ -1,33 +1,29 @@
...
-		message: 'HappenedAt must be a valid ISO 8601 date string'
-	})
-});
+export type CreateTeamData = z.infer<typeof CreateTeamSchema>;
+
+export const UpdateTeamSchema = CreateTeamSchema
+	.partial()
+	.refine(
+		(data) => Object.keys(data).length === 0,
+		{ message: 'At least one property is requires' }
```
lylaminju: typo: 'requires' -> 'required'

src/routes/team/index.ts
```diff
@@ -1,102 +1,271 @@
...
-		return context.json<StatusResponse>({ message: err?.message ?? 'An error has occurred' }, StatusCodes.INTERNAL_SERVER_ERROR);
-	}
-});
+			const isTeamIdValid = await validateExistingId(context.env.database, DBTables.TEAM, id);
 
-teamRoutes.patch('/:id', async (context: Context<EnvironmentBindings>) => {
-	try {
-		const { success: isValidTeamId, data: teamId } = IdSchema.safeParse(context.req.param('id'));
+			if (!isTeamIdValid) {
+				return context.json({ message: 'Invalid profile ID' } satisfies StatusResponse, StatusCodes.BAD_REQUEST);
```
lylaminju: typo: 'Invalid profile ID' -> 'Invalid team ID'

Actually, there are two more of the same typos in this file. You can check them!

lylaminju (commented):
You worked hard! 👍 Thank you for the kind explanation in PR template

Utilizing `openapi` is cool, but I’m a bit sorry that we have to include `satisfies` in every response. I should study this more.

Thanks to you, I learned HATEOAS for the first time. I want to see how useful it is in practice later.

migrations/0002_create-profile.sql
```diff
@@ -0,0 +1,67 @@
...
+	-- And should be unique inside the system
+	email TEXT NOT NULL UNIQUE,
+	-- Schema version to use
+	schemaVersion INTEGER NOT NULL DEFAULT 1,
+	-- The person's name, or how they like to be identified
+	name TEXT NOT NULL,
+	-- A text blurb the person can provide about themselves
+	description TEXT,
+	-- A flag indicating if the user is based on the Grater Toronto Area (GTA)
+	isBasedOnGTA INTEGER NOT NULL DEFAULT 1 CHECK(isBasedOnGTA IN (0, 1)),
```
manvinderjit, line 23: @madcampos Just nitpicking. Is there a remote possibility that TorontoJS will have physical events outside of GTA? In case the org wants to establish chapters in other areas? If no, then all good. If yes, then wouldn't having basedAt / location flags be better than isBasedOnGTA / canJoinLocalEvents? Or should we update the schema in the future as required ?

migrations/0002_create-profile.sql
lylaminju, line 1: @madcampos I think the consensus at the [slack thread](https://torontojs.slack.com/archives/C0805K3R8VB/p1737758214976389?thread_ts=1737752433.535969&cid=C0805K3R8VB) so far was to add `activatedAt` and `deletedAt` fields to profile.

| User Status                                    | activatedAt | deletedAt |
|------------------------------------------------|------------|-----------|
| Submit sign-up, not confirmed email yet       | null       | null      |
| After sign-up, confirmed email                | {{date}}   | null      |
| Left the community                            | {{date}}   | {{date}}  |
| Administrator deactivated (?) this user       | ?          | ?         |

<sub>👆please correct me if I'm wrong</sub>

But, just in case, do we have a scenario like deactivating the account for security reason(or something else) ?
I'm asking this because it's confusing to set those fields if deactivating user is possible

src/routes/profile/data.ts
madcampos: This is unfortunately an annoyance of zod. To use the [`omit`](https://zod.dev/?id=pickomit) utility it expects an object where all values are `true`... 🤷‍♂️

My first instinct was to have it as an array of strings with all the names, but zod started complaining 🫠

src/routes/profile/index.ts
madcampos: Yes, it makes more sense.

migrations/0002_create-profile.sql
madcampos: So, one example is the one and only Ken, that does not live in Toronto not even in Canada! 😅

The idea for these two flags is that we have a way to geographically locate people without asking for indentifiable information like the city they are in (with the `isBaseedOnGTA` flag) and also cover for situations that people that are in the GTA can't join in person events and are only available online (with `canJoinLocalEvents`).

migrations/0002_create-profile.sql
madcampos: Thank you for bringing this up! I had forgotten about this! Will add those fields to the schema.

I think there is no special case for deactivating an account, it would be only generally deactivating 🤷‍♂️

src/routes/team/index.ts
madcampos: Oops... That was copy + pasting on my part.

src/routes/profile/data.ts
madcampos: I made a slight change so it both the db fields and the values are abstracted away.

src/routes/profile/index.ts
madcampos: One argument for sending a `Bad Request` instead is that it is a "client error" in the sense that the id is incorrect and it was a value provided by the client. So I think in this case both are valid responses.

I'll change to `Not Found` though, because it makes more sense and was my original intent.

src/routes/profile/index.ts
madcampos: I think my initial reasoning was to add some permission, like, the user can or cannot update the data... But that was mixed up with a db error.

In this case, yes, `Internal Server Error` makes sense

migrations/0002_create-profile.sql
madcampos: I was reviewing the code and checking for SQLite data types and remembered _why_ it is a `TEXT` instead of `DATE`: it is supposed to be an ISO timestamp, not simply a `DATETIME` like in standard SQL.

I ran some tests and apparently ISO timestamps work out of the box with the `DATETIME` type in SQLite, so I've updated the types and added more context to the comments.

migrations/0002_create-profile.sql
lylaminju: @madcampos Okay, then how are the `activatedAt` and `deletedAt` fields set when a user is deactivated? Should we add a new field `deactivatedAt`?

migrations/0002_create-profile.sql
manvinderjit: Okay, makes sense if obscuring the location data is a priority. I guess we can update the schema with a flag in the future in case Ken decides to host an in-person event locally. :smile:

manvinderjit: You folks did put quite a bit of effort into this. :clap:

madcampos: Yeah, it would be great to have more granular data, but that would also mean a headache for privacy and security. So this bit is a good enough proxy that is not so bad to collect.

I'll add the reasoning to the docs

migrations/0002_create-profile.sql
madcampos: I was talking to @kbventures and `deletedAt` is maybe not the best name for this field. `deactivatedAt` or `inactivatedAt` makes more sense as this field serves to say "the profile is inactive and can't change anything on the platform".

src/routes/profile/data.ts
kbventures: @rjtbansal prepared statements so far has been very effective, sufficiently secure and easy to use even though most of the team is used to working with ORM's.

migrations/0002_create-profile.sql
kbventures: Ideally we don't need to add an additional field and just  meaningfully named multi purpose(context) properties.

migrations/0002_create-profile.sql
```diff
@@ -0,0 +1,87 @@
...
+	-- A profile with this flag set will not be able to login to the vms.
+	--
+	-- In case a user returns to the community and wants to reactivate their account,
+	-- that must be done manually by one of the organizers by removing this information.
+	--
+	-- In case a user is removed from the community, this flag is to be set, so their profile is deactivated.
+	--
+	-- In the future, we may use this flag as a potential "ban list" for spammers and similar situations.
+	-- If a person thinks it was mistakenly flagged as spam, then contatcing one of the organizers should resolve the issue.
+	deactivatedAt DATETIME DEFAULT NULL
```
madcampos: @pmjuu @kbventures I've made some changes, let me know what you think

lylaminju: It looks perfect! Thank you for the detailed explanation.

If there's a really trivial nitpick....
If users can be mistakenly flagged, providing a log of deactivation reasons could help organizers manage appeals. (adding a deactivationReason column or keeping a history table)

pmjuu: It looks perfect! Thank you for the detailed explanation.

If there's a really trivial nitpick....
If users can be mistakenly flagged, providing a log of deactivation reasons could help organizers manage appeals. (adding a deactivationReason column or keeping a history table)

src/index.ts
```diff
@@ -2,14 +2,22 @@ import { OpenAPIHono } from '@hono/zod-openapi';
...
 	defaultHook: statusResponseFormatter
 });
 
+// Catch all error handler.
+app.onError((err, context) => {
+	// TODO: add better error logging?
+	console.error(err);
+
+	return context.json({ message: 'An error has occured' }, StatusCodes.INTERNAL_SERVER_ERROR);
+});
```
lylaminju, line 19: @madcampos I'm not sure this method is recommended,, but, what about globally handling [SyntaxError containing 'JSON'](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Errors/JSON_bad_parse) to return BAD_REQUEST?
I think there is only invalid JSON format error when the SyntaxError message includes 'JSON'

```js
app.onError((err, context) => {
    if (err instanceof SyntaxError && err.message.includes('JSON')) {
        return context.json({ message: `Invalid JSON format: ${err.message}` }, StatusCodes.BAD_REQUEST);
    }

    console.error("Unhandled error:", err);
    return context.json({ message: 'An error has occurred' }, StatusCodes.INTERNAL_SERVER_ERROR);
});
```

Otherwise, we would have to handle SyntaxError while parsing the request body in every route that requires a JSON request body. (plus, handle ZodError in every route!)
```js
authRoutes.post('/sign-up', async (context: Context<EnvironmentBindings>) => {
   try {
    const body = await context.req.json();
    parsedBody = CreateProfileSchema.parse(body);
  } catch (error) {
    if (error instanceof SyntaxError) {
	return context.json<StatusResponse>({ message: `Invalid JSON format: ${error.message}` }, StatusCodes.BAD_REQUEST);
    }
    if (error instanceof ZodError) {
	return context.json<StatusResponse>({ message: `Invalid input: ${error.message}` }, StatusCodes.BAD_REQUEST);
    }
    console.error('Unexpected error parsing request body', error);
    throw error;
  }
```

lylaminju: or maybe use a new middleware?
```js
app.use('*', async (c, next) => {
  // Only attempt to parse JSON if the method supports a request body
  const methodsWithBody = ['POST', 'PUT', 'PATCH'];
  
  if (methodsWithBody.includes(c.req.method)) {
    try {
      const body = await c.req.json();
      c.req.parseBody = body; // Store parsed body in context
    } catch (err) {
      throw new JSONFormatError(err.message); // custom error
    }
  }
  
  await next();
});

// Global Error Handler
app.onError((err, c) => {
  if (err instanceof JSONFormatError) {
    return context.json<StatusResponse>({ message: `Invalid JSON format: ${err.message}` }, StatusCodes.BAD_REQUEST);
  }
  // .....
}
```
But, if feels a bit weird because I should check the http method.. `if (methodsWithBody.includes(c.req.method))`
Sorry if I'm being too verbose 😅

madcampos: So, there are a couple of things here...

With the openapi changes, parsing and validation of the request body is done for us. That is done in this part of the request handler:
```typescript
createRoute({
  request: {
    body: { content: { 'application/json': { schema: ValidationSchema } }, required: true }
  }
})
```

The `ValidationSchema` is a zod validator that will be used to validate the request body. And setting `required` to `true` makes so the body is [required and it needs to be in json format](https://github.com/honojs/middleware/tree/main/packages/zod-openapi#:~:text=If%20you%20want%20to%20force%20validation%20of%20requests%20that%20do%20not%20have%20the%20proper%20Content%2DType%2C%20set%20the%20value%20of%20request.body.required%20to%20true.), or else the request will fail.

Then, the handler for validation error responses is defined here: https://github.com/torontojs/vms/blob/5aa50f6ac52d2894860c17484ac1a18e46d69d57/src/utils/responses.ts#L212-L222

(And then it is applied to all routes for profile here (we unfortunately have to do it for every routes namespace): https://github.com/torontojs/vms/blob/5aa50f6ac52d2894860c17484ac1a18e46d69d57/src/routes/profile/index.ts#L19

I think yes, we can add the `BAD_REQUEST` return in case the json can't be parsed/syntax error here.

The [`app.onError` handler](https://hono.dev/docs/api/hono#error-handling) is specifcally for uncaught app errors. Like, "of dang, the app broke in a very weird and unexpected way".

We could also throw the [`HTTPException`](https://hono.dev/docs/api/exception#throw-httpexception) for some errors if we want and improve the `app.onError` to only handle errors that are not instances of `HTTPException`. (found out about this yesterday...)

Like, if it is an http exception, it is considered treated/handled and should be returned to the client, if it is any other error, then we catch and log it.

madcampos: I'm definitely bad at naming things... `statusResponseFormatter` should be renamed to something like `validationErrorFormatter`... 🫠

migrations/0002_create-profile.sql
```diff
@@ -0,0 +1,104 @@
...
+	-- In case a user returns to the community and wants to reactivate their account,
+	-- that must be done manually by one of the organizers by removing this information.
+	--
+	-- In case a user is removed from the community, this flag is to be set, so their profile is deactivated.
+	--
+	-- In the future, we may use this flag as a potential "ban list" for spammers and similar situations.
+	-- If a person thinks it was mistakenly flagged as spam, then contatcing one of the organizers should resolve the issue.
+	deactivatedAt DATETIME DEFAULT NULL,
+	-- When a profile is deactivated, this fields enables us to keep notes for other organizers in a future moment.
+	deactivatedReason TEXT DEFAULT NULL
```
kbventures: Missing a comma.

kbventures (commented):
Great work Marco!

kbventures (approved):

PR #89, opened 2025-02-26
Feat: roles route

lylaminju:
## Additional information

add `deletedAt` in `BaseDBEntity` as optional

### DIFF
--- a/migrations/0004_create-role.sql
+++ b/migrations/0004_create-role.sql
@@ -9,7 +9,7 @@ CREATE TABLE IF NOT EXISTS role (
 	-- Schema version to use
 	schemaVersion INTEGER NOT NULL DEFAULT 1,
 	-- The role name
-	role TEXT NOT NULL,
+	name TEXT NOT NULL,
 	-- The role description
 	description TEXT,
 	-- The UUID of the team this role belongs to
@@ -20,6 +20,8 @@ CREATE TABLE IF NOT EXISTS role (
 	happenedAt DATETIME NOT NULL,
 	-- The date this role was added to the database, saved as an ISO timestamp
 	insertedAt DATETIME NOT NULL,
+	-- The date this role was closed/deleted, saved as an ISO timestamp
+	deletedAt DATETIME DEFAULT NULL,
 
 	PRIMARY KEY (id),
 	FOREIGN KEY (teamId) REFERENCES team(id),
--- a/src/constants/db.ts
+++ b/src/constants/db.ts
@@ -5,7 +5,8 @@ export const SCHEMA_VERSION = 1;
 export enum DBTables {
 	PROFILE = 'profile',
 	TEAM = 'team',
-	PASSWORD = 'passwords'
+	PASSWORD = 'passwords',
+	ROLE = 'role'
 }
 
 export const IdAndSchemaVersionSchema = z.object({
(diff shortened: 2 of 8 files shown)


migrations/0004_create-role.sql
kbventures, line 1: What is the rational for using name instead of role?

src/constants/db.ts
kbventures, line 1: Why is the deleted  date optional?

kbventures (changes requested):

migrations/0004_create-role.sql
lylaminju: I just intended to follow the `team` schema pattern

src/constants/db.ts
lylaminju: If a new entry is created, there'll be no `deletedAt`

src/routes/profile/data.ts
```diff
@@ -23,7 +21,10 @@ export async function insertProfile(database: D1Database, { email, name, passwor
 			insertedAt,
 			email,
 			name,
-			description
+			description ?? '',
```
kbventures: Why are pronouns, birthday, avatar fields handled as '' and description ?? ''

src/routes/role/data.ts
```diff
@@ -0,0 +1,58 @@
+import { DBTables, generateBaseDBfields } from '../../constants/db.ts';
+import type { CreateRoleData, Role, UpdateRoleData } from './validation.ts';
+
+export async function insertRole(database: D1Database, data: CreateRoleData) {
+	const baseDbfields = generateBaseDBfields();
+
+	const { success } = await database.prepare(`
```
kbventures, line 7: This is using an older pattern that is less readable than the newest one:

https://github.com/torontojs/vms/blob/dev/src/routes/profile/data.ts

kbventures (changes requested):

src/routes/profile/data.ts
lylaminju: Oops it was my mistake! Thanks. I'll fix it

src/routes/role/data.ts
```diff
@@ -0,0 +1,58 @@
...
+		VALUES (
+			${[...Object.keys(baseDbfields)].fill('?').join(', ')},
+			${[...Object.keys(data)].fill('?').join(', ')}
+		)
+	`)
+		.bind(...Object.values(baseDbfields), ...Object.values(data))
+		.run();
+
+	return { success, id: baseDbfields.id };
+}
```
kbventures: This is using an older pattern that is less readable than the newest one:

https://github.com/torontojs/vms/blob/dev/src/routes/profile/data.ts

kbventures (changes requested):

src/routes/role/data.ts
lylaminju: Yes. It compromises readability. But on the bright side, I thought using `${Object.keys(baseDbfields).join(', ')},` pattern prevents typo and ensures consistency in INSERT queries.

IMO, pros outweigh the cons if we're used to using it.
We can always see the `baseDbfields` object properties when we hover over it :)

@kbventures What do you think?

kbventures: Readability is important because it makes the code easier to maintain, and it allows new developers to understand it more quickly. Additionally, it's essential that we align as a team on a standard, as this is what we have recently adopted as a group during a meeting but this is not a hill I'm willing to die on.

kbventures: I think ultimately the most important ting that we can get out of this is to align on a standard as a team.  Whether that is the version in the link of yours is less important.

kbventures (approved):

PR #90, opened 2025-02-26
Feat: auth middleware

kbventures:
## Summary

|  APIs \ Auth  |  Login  |  Access level  |
|-------------|--------|---------------|
|  **Auth** |  |  |
|  POST /auth/sign-up  |  -  |  -  |
|  GET  /auth/activate  |  -  |  -  |
|  POST /auth/sign-in  |  -  |  -  |
| **Profiles** |  |  |
| GET  /profiles  |  -  |  -  |
| GET /profiles/{id}  |  -  | -  |
| POST /profiles  |  O  |  `organizer`  |
| PATCH /profiles/{id}  |  O  |  `admin` or `volunteer` with own id  |
| DELETE /profiles/{id}  |  O  |  `admin`  |
| **Teams** |  |  |
| GET  /teams  |  O  |  `volunteer`  |
| GET /teams/{id}  |  O  |  `volunteer`  |
| POST /teams  |  O  |  `admin`  |
| PATCH /teams/{id}  |  O  |  `organizer`  |
| DELETE /teams/{id}  |  O  |  `admin`  |

## Additional information

- used the middleware property in [`createRoute()`](https://github.com/honojs/middleware/tree/main/packages/zod-openapi#configure-middleware-for-each-endpoint) of zod-openapi


src/middleware/auth.ts
```diff
@@ -0,0 +1,30 @@
...
+		if (new Date(sessionData.expiry) > new Date()) {
+			await context.env.SESSION_TOKENS.delete(sessionToken);
+			return context.json({ message: 'Session expired' }, StatusCodes.UNAUTHORIZED);
+		}
+
+		context.set('session', sessionData);
+		return next();
+	} catch (error) {
+		return context.json({ message: 'Invalid token' }, StatusCodes.INTERNAL_SERVER_ERROR);
+	}
```
evert: Don't catch generic errors. It's a a very bad pattern! Everyone in the VMS team needs to unlearn :grimacing: 

LGMT otherwise though!

kbventures:
Role-based service VS Seperate Routes

| Role-Based Service | - Centralized logic<br>- Easy to modify rules<br>- Good for complex permissions | - More initial setup<br>- More complex structure |

| Separate Routes | - Clear middleware chain<br>- Reusable role checks<br>- Easy to add new role-based routes | - More files to manage<br>- Need to maintain middleware separately |

The middleware approach (Option 2) is particularly powerful because:
Middleware is reusable across different routes
Role checks are separated from business logic
Easy to add new role-based routes
Clear separation of concerns

There are other options but I didn't really think single route was a good idea as we'd probably end up having to refactor it later.

migrations/0005_access.sql
```diff
@@ -0,0 +1,21 @@
...
+	-- This determines what actions they can perform
+	access TEXT NOT NULL CHECK(access IN ('admin', 'organizer', 'volunteer')),
+	-- The date this access was created
+	happenedAt DATETIME NOT NULL,
+	-- The date this access was added to the database
+	insertedAt DATETIME NOT NULL,
+
+	PRIMARY KEY (id),
+	FOREIGN KEY (id) REFERENCES profile(id)
+);
```
pmjuu, line 21: Using a primary key as a foreign key is not ideal because it enforces a strict one-to-one relationship, limiting flexibility. This prevents a profile from having multiple access roles.

I suggest using a separate `id` as the primary key while keeping `profile_id` as a foreign key.

```suggestion
CREATE TABLE IF NOT EXISTS access (
	-- Unique identifier for each access entry
	id TEXT NOT NULL UNIQUE COLLATE BINARY,
	-- The UUID of the profile this access belongs to
	profile_id TEXT NOT NULL UNIQUE COLLATE BINARY,
	-- Schema version to use
	schemaVersion INTEGER NOT NULL DEFAULT 1,
	-- The person's access level in the system
	-- This determines what actions they can perform
	access TEXT NOT NULL CHECK(access IN ('admin', 'organizer', 'volunteer')),
	-- The date this access was created, saved as an ISO timestamp
	happenedAt DATETIME NOT NULL,
	-- The date this access was added to the database, saved as an ISO timestamp
	insertedAt DATETIME NOT NULL,

	PRIMARY KEY (id),
	FOREIGN KEY (profile_id) REFERENCES profile(id)
);
```
For your understanding, you can also refer to the `0004_create-role.sql`

migrations/0006_password.sql
lylaminju, line 1: Actually, Marco added a table creation query for passwords. So we should remove the duplicate one.

https://github.com/torontojs/vms/pull/85/files#diff-09bf5067e985b3066a8b06e840d0dc0e9a621424c48e22fccccbd8f5c992840fR65-R76

src/types/data/session.d.ts
```diff
@@ -0,0 +1,5 @@
+export interface Session {
+	email: string;
+	role: Role;
+	expiry: string;
```
lylaminju: nitpick: if this is an ISO string, we can explicitly use `ISODate` type. It's just for conveying the meaning

https://github.com/torontojs/vms/blob/52c126e2c11d395395f83125a32d340843e1aaf7/src/types/global.d.ts#L11

src/routes/profile/data.ts
```diff
@@ -33,7 +33,15 @@ export async function insertProfile(database: D1Database, { email, name, passwor
...
-		`).bind(id, schemaVersion, password)
+		`).bind(id, schemaVersion, password),
+		database.prepare(`
+			INSERT INTO ${DBTables.ACCESS} (
+				id, access
+			)
+			VALUES (
+				?, ?
+			)
+		`).bind(id, 'volunteer')
```
pmjuu: We should add other columns like schemaVersion, insertedAt, happendAt

src/index.ts
```diff
@@ -54,11 +56,33 @@ app.doc('/open-api.json', {
...
-app.route('/teams', teamRoutes);
+// Public routes
 app.route('/auth', authRoutes);
-
 // Handle static assets using Cloudflare Workers
 app.get('/assets/*', async (context: Context<EnvironmentBindings>) => context.env.ASSETS.fetch(context.req.raw));
 
+// Protected routes - note hierarchy of access
+app.use(authMiddleware);
+app.route('/api')
```
lylaminju: Could you explain why you added new [grouping](https://hono.dev/docs/api/routing#grouping)?

src/routes/profile/data.ts
```diff
@@ -33,7 +33,15 @@ export async function insertProfile(database: D1Database, { email, name, passwor
...
-		`).bind(id, schemaVersion, password)
+		`).bind(id, schemaVersion, password),
+		database.prepare(`
+			INSERT INTO ${DBTables.ACCESS} (
+				id,access
+			)
+			VALUES (
+				?, ?
+			)
+		`).bind(id, 'volunteer')
```
lylaminju: We should add other columns like schemaVersion, insertedAt, happendAt

src/routes/auth/data.ts
```diff
@@ -3,7 +3,13 @@ import type { Profile, SignInData } from './validate.ts';
 
 export async function getPassword(database: D1Database, body: SignInData): Promise<string> {
 	const { results } = await database
-		.prepare(`SELECT password FROM ${DBTables.PROFILE} WHERE email = ? AND activatedAt IS NOT NULL LIMIT 1`)
+		.prepare(`
+			SELECT pw.password
+			FROM ${DBTables.PROFILE} pf
+			JOIN ${DBTables.PASSWORD} pw ON pf.id = pw.id
+			WHERE pf.email = ? AND pf.activatedAt IS NOT NULL
+			LIMIT 1
```
madcampos: **question:** To avoid a `JOIN`, which can be an expensive operation, would it make sense to move or duplicate the data on the `password` table?‏

src/routes/auth/index.ts
```diff
@@ -98,12 +98,11 @@ authRoutes.post('/sign-in', async (context: Context<EnvironmentBindings>) => {
 	}
 
 	const hashedPasswordWithSalt = await getPassword(context.env.database, parsedBody);
-	const profileId = await getId(context.env.database, parsedBody);
-
 	if (!hashedPasswordWithSalt) {
 		return context.json<StatusResponse>({ message: 'Invalid email or password' }, StatusCodes.UNAUTHORIZED);
 	}
 
+	const profileId = await getId(context.env.database, parsedBody);
```
madcampos: **nitpick:** Please rename this function to `getProfileId` or something more explicit.‏

`getId` imediatly brings the question "id _for what_?".

Adding a specifier here will help removing ambiguty and mental overhead.

migrations/0005_access.sql
```diff
@@ -0,0 +1,19 @@
+-- Migration number: 0005 	 2025-01-30T00:58:45.123Z
+
+DROP TABLE IF EXISTS access;
+
+-- Store authentication and authorization data separately from profiles
+CREATE TABLE IF NOT EXISTS access (
```
lylaminju, line 6: Having `access` column in `access` table sounds a bit weird to me. How about renaming them?

For example, `access` column in `auth` table, or `level` column in `access` table, or something else..

src/routes/auth/data.ts
```diff
@@ -3,7 +3,15 @@ import type { Profile, SignInData } from './validate.ts';
...
 	const { results } = await database
-		.prepare(`SELECT password FROM ${DBTables.PROFILE} WHERE email = ? AND activatedAt IS NOT NULL LIMIT 1`)
+		.prepare(`
+            SELECT password
+            FROM ${DBTables.ACCESS}
+            WHERE id IN (
+                SELECT id
+                FROM ${DBTables.PROFILE}
+                WHERE email = ? AND activatedAt IS NOT NULL
+            )
```
lylaminju: I recommend using a JOIN query instead of a nested subquery. JOIN explicitly defines the relationship between tables, which enhances clarity and may optimize query performance. This approach can improve both readability and efficiency.
And,, don't forget to set LIMIT 1

```suggestion
            SELECT a.password
            FROM ${DBTables.PROFILE} p
            JOIN ${DBTables.ACCESS} a ON a.id = p.id
            WHERE p.email = ? AND p.activatedAt IS NOT NULL
            LIMIT 1
```
ref)
https://stackoverflow.com/questions/2577174/join-vs-sub-query
https://www.geeksforgeeks.org/sql-join-vs-subquery/

src/middleware/createMiddleware.ts
lylaminju, line 1: We can utilize enum instead of writing raw strings like 'admin', 'organizer', 'volunteer'

src/routes/auth/data.ts
pmjuu: @madcampos Do you mean to store the password in `profile` table?

migrations/0005_access.sql
madcampos: Yes to renaming!

I can't think of a better name than `access` to the table. But we can try to come up with a name for "access level, password, and internal stuff".

`Auth` seems a little bit off though... It _is_ for authorization, but not only that

src/routes/auth/data.ts
madcampos: Yeah, no worries, but my question is more like: why don't we keep all of the data in one table?

Like:
```sql
CREATE TABLE IF NOT EXISTS access (
	id TEXT NOT NULL UNIQUE COLLATE BINARY,
	schemaVersion INTEGER NOT NULL DEFAULT 1,
	access TEXT NOT NULL CHECK(access IN ('admin', 'organizer', 'volunteer')),
	password TEXT NOT NULL,

	email TEXT NOT NULL UNIQUE, -- <= Add it here
	activatedAt DATETIME DEFAULT NULL, -- <= And here

	PRIMARY KEY (id),
	FOREIGN KEY (id) REFERENCES profile(id)
);
```

So the query would only look for one table instead of multiple. It _does_ duplicate information, but may be interesting for performance.

The argument here is as this is the auth endpoint and a verification we will need to do multiple times, potentially on every request (as part of the auth middleware), we would benefit from a _very fast_ query.

lylaminju: Sounds reasonable! yeah, there'll be a lot of requests to get password, so we would benefit from a fast query.

src/index.ts
```diff
@@ -54,11 +56,29 @@ app.doc('/open-api.json', {
...
+		if (requestedId !== session.id) {
+			return c.json({ message: 'Forbidden: Can only update own profile' }, StatusCodes.FORBIDDEN);
+		}
+
+		return profileRoutes.fetch(c.req.raw);
+	});
+
+app.route('/teams')
+	.get('/', async (c) => teamRoutes.fetch(c.req.raw))
+	.get('/:id', async (c) => teamRoutes.fetch(c.req.raw));
```
lylaminju: 1. There are some missing endpoints like deleting profile, creating team,, etc
2. How about configuring middlewares inside `createRoute()`? 
```js
profileRoutes.openapi(
	createRoute({
		method: 'post',
		// ......
		middleware: [authMiddleware, authorizeOrganizer] as const // 👈 HERE!
	}),
	async (context) => {
		const body = context.req.valid('json');
	// .......
```
please check this [example](https://github.com/honojs/middleware/tree/main/packages/zod-openapi#configure-middleware-for-each-endpoint) for the reference

src/routes/auth/data.ts
kbventures: I believe we decided in the last meeting to have a non timeline tracked table including id has foreign key, schema version, password and access.  Are you referring to something else?

migrations/0005_access.sql
kbventures: Alternatives to access table: 

Credentials
Security
Auth
Identity
Login.  

We could use several words but that would break from our current convention.  For example: 

user_authentication
user_access_credentials
user_security_info

 We can keep access for the table name and the actual column could be named authorization?

src/routes/auth/data.ts
kbventures: According to the geeks for geeks article utilizing a join is bad for readability but is better for performance.   I opted for sub query for readability and assumed performance wouldn't be an issue due to the volume expected from the platform.

migrations/0005_access.sql
kbventures: I think we can keep access for the table name and the actual column could be named authorization?

src/index.ts
kbventures: Draft of this solution implemented but not complete yet.

src/routes/profile/index.ts
```diff
@@ -15,18 +17,101 @@ import { IdParamSchema, validateExistingId } from '../../utils/validation.ts';
...
+// POST new profile
+protectedProfileRoutes.openapi(
 	createRoute({
 		method: 'post',
 		path: '/',
 		operationId: 'createNewProfile',
 		summary: 'Create new profile',
 		description: 'Add a new profile to the VMS including basic information about this person.',
 		tags: ['Profile'],
+		middleware: [authMiddleware, authorizationVolunteer] as const,
```
lylaminju: Since you already added `app.use('/*', authMiddleware);` in `index.ts`
https://github.com/torontojs/vms/pull/90/files#diff-a2a171449d862fe29692ce031981047d7ab755ae7f84c707aef80701b3ea0c80R68

I think you don't have to put `authMiddleware` here one more time.

migrations/0005_access.sql
madcampos: I like `Credentials` and `Identity` for table name.

For the column name, I think it whould be `accessLevel` instead.

It is not about authorization as it doesn't control specifically if the user is authorized to do things or not on the platform (that is like the password and all).

And it also disambiguates from `role` or `accessRole`.

src/index.ts
```diff
@@ -54,11 +54,17 @@ app.doc('/open-api.json', {
...
 // Handle static assets using Cloudflare Workers
 app.get('/assets/*', async (context: Context<EnvironmentBindings>) => context.env.ASSETS.fetch(context.req.raw));
 
+// Public routes
+app.route('/profiles', publicProfileRoutes);
+
+// Protected routes (after auth middleware)
+app.use('/*', authMiddleware);
+app.route('/profiles', protectedProfileRoutes);
+app.route('/teams', teamRoutes);
```
madcampos:
```suggestion
app.route('/profiles', publicProfileRoutes);
app.route('/profiles', authMiddleware, protectedProfileRoutes);

app.route('/teams', authMiddleware, publicTeamRoutes);
app.route('/teams', authMiddleware, protectedTeamRoutes);
```

Being very nitpicky here: I'd rather we explicitly use middlewares than use it for all routes. It makes it easier to reason about the code.

src/middleware/auth.ts
```diff
@@ -0,0 +1,26 @@
...
+	if (!sessionToken) {
+		return context.json({ message: 'Invalid or missing token' }, StatusCodes.UNAUTHORIZED);
+	}
+	const sessionData: SessionData | undefined = await context.env.SESSION_TOKENS.get(sessionToken, 'json');
+
+	if (!sessionData) {
+		return context.json({ message: 'Invalid session' }, StatusCodes.UNAUTHORIZED);
+	}
+
+	if (new Date(sessionData.expiry) > new Date()) {
```
madcampos: Isn't that the other way arround?

If the expiry date is before (`<`) the current date, it means the token is expired.

If the expiry date is after (`>`) the current date, it means the expiry date haven't happened yet.

src/middleware/createMiddleware.ts
```diff
@@ -0,0 +1,44 @@
...
+		return context.json({ message: 'Forbidden' }, StatusCodes.FORBIDDEN);
+	}
+	return next();
+};
+
+export const authorizeAdmin = createAccessMiddleware(Access.ADMIN);
+export const authorizeOrganizer = createAccessMiddleware(Access.ORGANIZER);
+export const authorizeVolunteer = createAccessMiddleware(Access.VOLUNTEER);
+
+export const canModifyOwnProfile = async (context: Context, next: Next) => {
```
madcampos: I think this logic should be moved to the profile files as it is very specific there.

And also rename the function as `canModifyProfile`. the way it is right now, the name is confusing as admins are allowed to change other people's profiles, but the name implies the check is only for the user's own profile.

src/middleware/createMiddleware.ts
```diff
@@ -0,0 +1,44 @@
...
+	const session = context.get('session') as SessionData;
+	const targetId = context.req.param('id');
+
+	// If admin or organizer, allow
+	if (session.access === Access.ADMIN || session.access === Access.ORGANIZER) {
```
madcampos: **question**: Should organizers be allowed to edit people's profiles?

In my opinion, no. The user can already edit most of their profile by themselves, and admins can edit the other fields, so there is no need for organizers to do so.

migrations/0005_access.sql
```diff
@@ -0,0 +1,23 @@
...
+	id TEXT NOT NULL UNIQUE COLLATE BINARY,
+	-- Schema version to use
+	schemaVersion INTEGER NOT NULL DEFAULT 1,
+	-- The person's access level in the system
+	-- This determines what actions they can perform
+	access_level TEXT NOT NULL CHECK(access IN ('admin', 'organizer', 'volunteer')),
+	-- The person's account password
+	password TEXT NOT NULL,
+	-- The person's email address
+	email TEXT NOT NULL UNIQUE,
```
madcampos, line 17: **question**: should we add `email` as a FK to `profile`?

src/routes/auth/data.ts
```diff
@@ -1,19 +1,43 @@
...
-		.prepare(`SELECT password FROM ${DBTables.PROFILE} WHERE email = ? AND activatedAt IS NOT NULL LIMIT 1`)
-		.bind(body.email)
+		.prepare(`
+            SELECT password
+            FROM ${DBTables.ACCESS}
+            WHERE email = ? AND activatedAt IS NOT NULL
+			LIMIT 1
+        `)
+		.bind(email)
+		.run<string>();
```
madcampos: Double check if it returns a plain string or an object with the `password` field.

src/routes/auth/data.ts
```diff
@@ -1,19 +1,43 @@
...
+
+	const password = results[0];
```
madcampos: **nitpick**: use destructuring
```suggestion
	const [password] = results;
```

src/routes/auth/data.ts
```diff
@@ -1,19 +1,43 @@
...
+		.bind(email)
 		.run<Profile>();
 
-	const password = results[0]?.password;
+	const id = results[0]?.id;
 
-	if (!password) {
-		throw new Error('No active account found with the provided email');
-	}
+	return id;
```
madcampos: **nitpick**: avoid the extra variable, it can be shortcut to:
```
return resutls[0]?.id;
```

src/routes/auth/data.ts
```diff
@@ -1,19 +1,43 @@
...
-	return password;
+export async function getAccessLevel(database: D1Database, profileId: string) {
+	const { results } = await database
+		.prepare(`SELECT access_level FROM ${DBTables.ACCESS} WHERE id = ? LIMIT 1`)
+		.bind(profileId)
+		.run<Access>();
+
+	const accessLevel = results[0];
+
+	return accessLevel;
```
madcampos: **nitpick**: avoid extra variable
```suggestion
	return results[0];
```

src/routes/auth/data.ts
```diff
@@ -1,19 +1,43 @@
...
-	}
+	return id;
+}
 
-	return password;
+export async function getAccessLevel(database: D1Database, profileId: string) {
+	const { results } = await database
+		.prepare(`SELECT access_level FROM ${DBTables.ACCESS} WHERE id = ? LIMIT 1`)
+		.bind(profileId)
+		.run<Access>();
```
madcampos: Double check here as well if the result is not actually an object with the `access_level` property

src/routes/auth/data.ts
```diff
@@ -27,10 +51,12 @@ export async function checkEmail(database: D1Database, email: string) {
...
-		.prepare(`UPDATE ${DBTables.PROFILE} SET activatedAt = ? WHERE email = ?`)
-		.bind(now, email)
-		.run();
-
-	return success;
+	const results = await database.batch([
+		database.prepare(`UPDATE ${DBTables.PROFILE} SET activatedAt = ? WHERE email = ?`)
+			.bind(now, email),
+		database.prepare(`UPDATE ${DBTables.ACCESS} SET activatedAt = ? WHERE email = ?`)
+			.bind(now, email)
```
madcampos: I completelly forgot, so please remind me again: why are we keeping the `activatedAt` field on both `profile` and `access_level`?

src/routes/auth/index.ts
```diff
@@ -97,11 +101,21 @@ authRoutes.post('/sign-in', async (context: Context<EnvironmentBindings>) => {
...
 
+	const profileId = await getProfileId(context.env.database, parsedBody.email);
+	if (!profileId) {
+		return context.json<StatusResponse>({ message: 'Invalid id or Account not activated' }, StatusCodes.UNAUTHORIZED);
+	}
+
+	const accessLevel = await getAccessLevel(context.env.database, profileId);
+	if (!accessLevel) {
+		return context.json<StatusResponse>({ message: 'Access not found' }, StatusCodes.NOT_FOUND);
+	}
```
madcampos, line 118: **question**: Why are we checking the access level here? Shouldn't all users be able to login?

If that function also checks for activation of profile, then here we should _only_ check for activation and update `getAccessLevel` to reference the activation check.

src/routes/auth/index.ts
```diff
@@ -97,11 +101,21 @@ authRoutes.post('/sign-in', async (context: Context<EnvironmentBindings>) => {
...
-	const hashedPasswordWithSalt = await getPassword(context.env.database, parsedBody);
+	const hashedPasswordWithSalt = await getPassword(context.env.database, parsedBody.email);
 	if (!hashedPasswordWithSalt) {
 		return context.json<StatusResponse>({ message: 'Invalid email or password' }, StatusCodes.UNAUTHORIZED);
 	}
 
+	const profileId = await getProfileId(context.env.database, parsedBody.email);
+	if (!profileId) {
+		return context.json<StatusResponse>({ message: 'Invalid id or Account not activated' }, StatusCodes.UNAUTHORIZED);
+	}
```
madcampos: **suggestion**: Here, instead of getting the profile id from the `profiles` table, we should instead:
1. Get both `profileId` and `password` from the `getPassword` (update the name to also include the profile id)
2. Check if the profile exists (or fetch the profile) on this second step that matches the `profileId` returned from the previous call.

src/routes/auth/index.ts
```diff
@@ -110,10 +124,13 @@ authRoutes.post('/sign-in', async (context: Context<EnvironmentBindings>) => {
 	const sessionToken = crypto.randomUUID();
 	const hoursAhead = 1;
 	const tokenExpiryISO = new Date(Date.now() + hoursAhead * 60 * 60 * 1000).toISOString();
```
madcampos: **nitpick**: I'd rather we change this in a couple ways:
- Do not do time math directly, it is very error prone and can make things error out badly (e.g. time zones, DST, adding time when it is close to a day change, leap seconds, leap years, etc)
- Use a constant instead of magic numbers. The "`60 * 60 * 1000`" part is not very intuitive, so it is better to use a constant like `EXPIRE_TIME_TO_ADD` or something (I'm bad with names).

src/routes/auth/index.ts
```diff
@@ -110,10 +124,13 @@ authRoutes.post('/sign-in', async (context: Context<EnvironmentBindings>) => {
...
-		email: parsedBody.email
-	});
+	const sessionDataObject: SessionData = {
+		id: profileId,
+		email: parsedBody.email,
+		access: accessLevel,
+		expiry: tokenExpiryISO
+	};
+	const sessionData = JSON.stringify(sessionDataObject);
 	await context.env.SESSION_TOKENS.put(sessionToken, sessionData);
```
madcampos, line 135: **question**: What happens if the user is already signed-in and have a session saved and they call this endpoint again?

src/routes/profile/data.ts
```diff
@@ -27,13 +27,34 @@ export async function insertProfile(database: D1Database, { email, name, passwor
...
+			)
+		`).bind(
+			id,
+			schemaVersion,
+			happenedAt,
+			insertedAt,
+			'volunteer',
+			'',
+			'',
+			id
```
madcampos: **nitpick:** this looks like a prettier formatting thing... Can we instead break it by blocks, like the prepared statement?‏

```suggestion
			id, schemaVersion, happenedAt, insertedAt,
			'volunteer', '', '', id
```

src/routes/profile/data.ts
```diff
@@ -27,13 +27,34 @@ export async function insertProfile(database: D1Database, { email, name, passwor
...
+		)
+		// TODO : set TorontoJS team id 👆
```
madcampos: Just generate a random UUID and save it on `constants.ts`, we don't need to make a big deal out of it 😅

src/types/data/session.d.ts
madcampos, line 1: **suggestion**: what about we move this to `access` instead?

src/types/data/authMiddleware.d.ts
madcampos, line 1: **suggestion**: what about moving this type closer to where it is used, like `auth.ts`?

src/types/data/access.ts
madcampos, line 1: **question**: what about we move this closer to where it is used?

src/index.ts
lylaminju: Yes, it would be better to reason. But, the `route()` requires max 2 parameters, so we can't use middleware inside `route()`. I don't know how to explicitly put middleware without `app.use('/*', authMiddleware);` 🤔 
Any ideas? @kbventures @bcha92

src/middleware/auth.ts
lylaminju: Oh, I think it's a mistake. Thank you!

src/middleware/createMiddleware.ts
lylaminju: It makes sense. Only admin and the volunteer himself can edit.

src/index.ts
bcha92: Seems like Hono doesn't allow a third argument for `app.route()`.
However, it recommends using middleware above the handler:


If you have the middleware that you want to execute, write the code above the handler.
```jsx

app.use(logger())
app.get('/foo', (c) => c.text('foo'))
```

Hono doc [here](https://hono.dev/docs/api/routing#routing-priority).

bcha92: Seems like for routing in Hono, handlers and middleware are executed in registration order.
There is no third argument for `.route()`, but the `.use()` usage for middleware is a bit vague and is only mentioned sparingly in this [page](https://hono.dev/docs/api/routing#routing-priority).

I did think there was a way to squeeze a third argument, but it only works for some frameworks like Express. Sadly Hono isn't one of them.

lylaminju: Well, if it would be impossible to explicitly use authMiddleware at `index.ts`, we should put authMiddleware in the `createRoute()` middleware property at `routes/profile/index.ts`

src/routes/auth/data.ts
lylaminju: I think we're keeping the `activatedAt` because we need to check if the access is valid in `getPassword()` query
https://github.com/torontojs/vms/pull/90/files#diff-94bdeaacf030d0247bc567e514e7ecdc13ce2fb014f0dff219bf42eef2b2b134R7-R10

But, there might be a better way not to keep the field on both `profile` and `access`

lylaminju: Since we can check the `activatedAt` from the `profile` table by calling `getProfileId()`, which checks `activatedAt`, before calling getPassword(), I think it'd better to remove `activatedAt` from the `access` table.

src/routes/auth/index.ts
lylaminju: @madcampos I intended to retrieve `access_level` to put it in the session data, not for checking purpose. But, do you think it's still unnecessary?

https://github.com/torontojs/vms/pull/90/commits/8c820d09ec3ea197b3e9830888f47a2483f66214#diff-b848d07b6c12d758efa86f57a831156d428e18118a886dde200312c99e4c0fddR133-R140

src/routes/auth/index.ts
lylaminju: Good point! I'm going to use `date-fns` to calculate time

src/routes/auth/index.ts
lylaminju: Good point! There might be some potential issues.

- Security vulnerability
  - Old session tokens remain in `SESSION_TOKENS` after a new token is issued, potentially allowing unauthorized access if stolen
- Resource waste
  - Unused old tokens accumulate in `SESSION_TOKENS` until they expire, increasing storage and management overhead.
- User Experience Concern
  - There is no notification or confirmation when a new session overwrites an existing one. This may confuse users.

So,, for the solution, we should check if an active session exists before issuing a new token. 
If the active session exists, the api should return message that notice the use is already signed-in.

Did I miss anything? Any ideas? @madcampos @kbventures @bcha92

bcha92: I propose one of two solutions (but other solutions are also welcome) in regards to active sessions:
1. User can only have one active session at a time and should clear once it is expired or logged out (do we have one for closed windows/tabs?).
2. Users should be allowed to register tokens on multiple devices unique to the device (i.e. PC/mobile) with timed session tokens issued when logging to in (similar to "approved devices" for faster verification) and should expire if unused/not logged in after X number of days.

...anything else, just a thought to get something off the ground.

src/routes/auth/index.ts
bcha92: Date as a constant should be a consistent value (try looking into `epoch`) which is a integer/number that doesn't take account of timezones. Easier to calculate, and is also convertible back into a date/time, then apply timezone (default timezone is UTC Greenwich).

src/index.ts
madcampos: Yeah, I thought Hono by itself would allow for an array of middlewares like express (that is actually a neat feature... oh well...)

But, the openapi thing we are using does [support an array of middlewares](https://github.com/honojs/middleware/tree/main/packages/zod-openapi#configure-middleware-for-each-endpoint), so maybe migrate to that soon-ish? 😅

src/routes/auth/index.ts
madcampos: I think we should be mindful here of DB queries. This piece of code runs for every single request, so it should be very optimized.

One option would be to do the query first and then use it's result as parameters for all the checks. That way we only query the database once and decouple checking for password and access from querying the database.

src/routes/auth/index.ts
madcampos: I like auto expire after 90 days or something like that.

But I also think that sending a request to sign-in again if you already have a valid token should just refresh the token with a new one. After all, you are signing-in again with all the credentials provided.

Another question is: how to verify the user device?

IP address for example, is unreliable, specially on mobile (moving cell towers, rebooting your phone, etc. may change your IP address)

madcampos (dismissed):
3 threads left to resolve, otherwise looks good to me.

**NOTE**: All of these _are non-blocking and/or require more discussion_, so I'm approving this PR for us to move forward and resolve those issues separately.

- [multiple middlewares](https://github.com/torontojs/vms/pull/90#discussion_r1994126949)
- [re-request sign-in](https://github.com/torontojs/vms/pull/90#discussion_r1994176163)
- [multiple db queries on auth middleware](https://github.com/torontojs/vms/pull/90#discussion_r1994150966)

src/index.ts
kbventures: @lylaminju @madcampos  the max two parameters is why I went down that route.

src/routes/auth/index.ts
kbventures: Discussion moved to [Handling Active Sessions and Token Refresh on Re-sign-in](https://github.com/torontojs/vms/issues/94)

src/routes/auth/index.ts
kbventures: Issue created: https://github.com/torontojs/vms/issues/95

src/index.ts
kbventures: Conversation moved to https://github.com/torontojs/vms/issues/96

src/constants/db.ts
```diff
@@ -48,7 +48,7 @@ export function generateBaseDBfields() {
 }
 
 export interface AccessSchema extends IdAndSchemaVersion {
-	access_level: Access;
```
madcampos: There is a mismatch between the property name in typescript and the name saved on the database.

src/routes/auth/data.ts
```diff
@@ -1,18 +1,18 @@
 import { type AccessSchema, DBTables } from '../../constants/db.ts';
 import type { Profile } from './validate.ts';
 
-export async function getProfileIdPassword(database: D1Database, email: string) {
+export async function getAccess(database: D1Database, email: string) {
 	const { results } = await database
 		.prepare(`
-            SELECT id, password
+            SELECT id, password, access_level
```
madcampos: Same here

src/routes/auth/index.ts
```diff
@@ -102,21 +102,17 @@ authRoutes.post('/sign-in', async (context: Context<EnvironmentBindings>) => {
 		throw error;
 	}
 
-	const { id: profileId, password: storedPassword } = await getProfileIdPassword(context.env.database, parsedBody.email);
-	if (!profileId || !storedPassword) {
+	const accessData = await getAccess(context.env.database, parsedBody.email);
+	if (!accessData) {
 		return context.json<StatusResponse>({ message: 'Invalid email' }, StatusCodes.UNAUTHORIZED);
 	}
+	const { id: profileId, password: storedPassword, accessLevel } = accessData;
```
madcampos: Please keep it destructured, there is no need to save to a variable to then destructure. And use as a check a field that should always be returned, like `profileId`

madcampos (approved):

PR #100, opened 2025-03-22
Fix: Add generic error messages to auth routes and create health-check route

### LINKED ISSUE #99: Error messages on login and password reset should all be the same
For security reasons, the error messages when a user tires to login should all be generic, and not specify which field is wrong.

That way, it is harder for attackers to guess emails and passwords.

On the case of password reset, even if the email is not on the platform, the request should return a success and fail silently to the client so it doesn't provide information about which emails are on the platform.

abdulqshabbir:
## Summary

- added a generic unauthorized message to sign-in route
- added a `health-check` route to check status of environment varirables

## Tests
- called api and verified I received generic error message in sign-in route
- tested health-check endpoint

## Additional information
- also added a similar generic message to the auth middleware (can remove if not ok)
- fail fast and throw error if send grid api key or send grid email not provided as env variables 

Reverted: capture error message in catch all handler and return to client to not return sensitive information back to client

Resolves #99


abdulqshabbir:
Will create a separate PR for doc changes relating to setting up local environment and send grid API key

src/routes/auth/index.ts
```diff
@@ -13,6 +13,13 @@ import { type SignInData, SignInSchema } from './validate.ts';
 export const authRoutes = new Hono();
 
 authRoutes.post('/sign-up', async (context: Context<EnvironmentBindings>) => {
+	if (!context.env.SENDGRID_API_KEY) {
```
madcampos: **suggestion:** I think this check can be moved to a util like `checkEnvVars` and then we add it to the root `index.ts` tile.

Like, when the server starts we check for all the env vars, and if there is any variable missing we fail the server from starting giving an earlier feedback

madcampos (dismissed):
Awesome PR!

Just one comment, but it is not a blocking issue for me, we can resolve that later.

src/routes/auth/index.ts
abdulqshabbir: I definitely agree! Will make the update on this PR later. I do like the idea of having a centralized check happening and the server should fail to start if the variables aren't present.

evert: Another nice spin on this, make that `env.ts` file also re-export environment variables, so you can guarantee the correct spelling with types

abdulqshabbir: Yeah totally agree! In our hono setup we access environment variables through a context object which is already typed making it easy to catch spelling errors. Im working on making sure that there's a run-time check for these env though at start up

abdulqshabbir:
@madcampos @evert

I tried to add a run-time check before the server starts but unfortunately ran into two problems: 
1.  Loading of `.dev.vars` secrets into `process.env` would require another dependency `dotenv` (I know how much @madcampos LOVES new dependencies). I know that node can read in env files without dot env but I'm not sure if that would work in a cloudflare environemnt. 
2. Bigger problem: the `wrangler.toml` file which supplies some of the variables (like `BASE_URL`) to `context.env` can only be read from within a hono request handler (unless I do some kind of toml parsing of the `wrangler.toml` file)

The solution for now (which is not great) is to run a check in middleware (where we have access to `context.env`and only perform this check in a `local` environment for now) and throw an error to the user. If any new variables are declared in `EnvironmentBindings` the developer will be forced to add an error for that missing env because of the type linking

If we keep all our env variables in one location `.dev.vars` and use something like `dotenv` we could do a runtime check for the variables. I think the issue is mostly around using some varirables from `wrangler.toml` and some from `.dev.vars`

src/routes/auth/index.ts
madcampos: @evert by "reexporting" the environment variables you mean we would only reference variables that are in `env.ts`?

madcampos:
@abdulqshabbir

> Loading of .dev.vars secrets into process.env would require another dependency dotenv (I know how much @madcampos LOVES new dependencies). I know that node can read in env files without dot env but I'm not sure if that would work in a cloudflare environemnt.

It won't work on cludflare because it uses workerd instead of node. They are similar but different.

> If we keep all our env variables in one location .dev.vars and use something like dotenv we could do a runtime check for the variables. I think the issue is mostly around using some varirables from wrangler.toml and some from .dev.vars

We can't rely on the `.dev.vars` file as it only exists and is used when running the server locally. On cloudflare the variables get injected into the running process.

> The solution for now (which is not great) is to run a check in middleware (where we have access to context.envand only perform this check in a local environment for now) and throw an error to the user. If any new variables are declared in EnvironmentBindings the developer will be forced to add an error for that missing env because of the type linking

I think we can have a solution that is similar to what you are describing, but instead of middlewares we have an endpoint. In my previous work we had a `/healthcheck` endpoint that returned general backend health information, like if things were set properly and if the server was up and running. We can have a similar endpoint and make it so more information is returned on a dev environment.

abdulqshabbir:
> It won't work on cludflare because it uses workerd instead of node. They are similar but different.

Ok makes sense, I'm not too familiar with the differences will read more about it

>We can't rely on the .dev.vars file as it only exists and is used when running the server locally. On cloudflare the variables get injected into the running process.

Yeah I think it would only be a local environment check in this case which would not be great. 

>I think we can have a solution that is similar to what you are describing, but instead of middlewares we have an endpoint. In my previous work we had a /healthcheck endpoint that returned general backend health information, like if things were set properly and if the server was up and running. We can have a similar endpoint and make it so more information is returned on a dev environment.

Ok sounds good! I think this is a good middle ground. The check won't have to run in middle-ware every time a request is made but still gives developers and opportunity to check if they're missing anything in their environment.

abdulqshabbir:
@madcampos I've updated the PR to add a `health-check` endpoint. It will check all environment variables that are available in `context.env` and using `envSchema` it will validate the env variables through zod. When a developer adds a new environment variable they have the flexibility to mark it as required/optional and any other validation they'd like to perform.

For production I've hidden away some of the sensitive details and returns a generic OK or NOT-OK.

Let me know what you think!

.gitignore
```diff
@@ -34,3 +34,6 @@ certs
 
 # Typescript build info
 *.tsbuildinfo
+
+# Bruno: Abdul testing out bruno
+bruno/
```
madcampos: [We don't talk about Bruno 🎵](https://www.youtube.com/watch?v=bvWRMAU6V-c)

src/middleware/auth.ts
```diff
@@ -6,19 +6,20 @@ import { StatusCodes } from '../utils/responses.ts';
 export const authMiddleware = async (context: Context, next: Next) => {
 	// Get token from cookie
 	const sessionToken: string | undefined = getCookie(context, 'auth_token');
+	const genericUnauthorizedMessage = 'Invalid or missing token';
```
madcampos: **suggestion:** In this case, we are returning the same response every time, why not save the response to a constant instead of only the message?‏

src/routes/auth/index.ts
```diff
@@ -102,24 +102,26 @@ authRoutes.post('/sign-in', async (context: Context<EnvironmentBindings>) => {
 		throw error;
 	}
 
+	const genericErrorMessage = 'The email and/or password is invalid or the account is not activated';
```
madcampos: **suggestion:** Maybe the same here?‏

src/routes/health-check/index.ts
```diff
@@ -0,0 +1,85 @@
...
+			},
+			[StatusCodes.UNPROCESSABLE_CONTENT]: {
+				description: 'Error response',
+				content: { 'application/json': { schema: StatusResponseSchema } }
+			}
+		},
+		middleware: [authMiddleware] as const
+	}),
+	(context: Context<EnvironmentBindings>) => {
+		type EnvToCheck = Omit<
```
madcampos: **thought:** Instead of using `Omit` here, why not change the global definition, so `EnvVars` is only for environment variables? Then the reference here is just to the variable, no type shennanigans.‏

src/routes/health-check/index.ts
```diff
@@ -0,0 +1,85 @@
...
+				SENDER_EMAIL: z
+					.string({ message: 'required for emails to be sent' })
+					.email({ message: 'must be a valid email address' }),
+				SENDGRID_API_KEY: z
+					.string({ message: 'required for emails to be sent' })
+					.min(1, { message: 'required for emails to be sent' }),
+				NODE_ENV: z
+					.enum(['development', 'local', 'production'], { message: 'must be one of the following: development, local, production' })
+			} satisfies Record<EnvKeys, ZodSchema>
+		);
```
madcampos: **nitpick:** Let's extract the schema and the function to check the env vars to `validation.ts`‏, that way it follows the same pattern as the other routes.

src/types/global.d.ts
```diff
@@ -1,13 +1,18 @@
...
-	Bindings: {
-		database: D1Database,
-		SESSION_TOKENS: KVNamespace,
-		ACTIVATION_TOKENS: KVNamespace,
-		BASE_URL: string,
-		SENDER_EMAIL: string,
-		SENDGRID_API_KEY: string,
-		ASSETS: Fetcher
-	};
+	Bindings: EnvVars;
```
madcampos: Basically this:
```typescript
interface EnvVars {
	BASE_URL: string;
	SENDER_EMAIL: string;
	SENDGRID_API_KEY: string;
	NODE_ENV: NodeEnvironment;
}

interface EnvironmentBindings {
	Bindings: {
		database: D1Database;
		SESSION_TOKENS: KVNamespace;
		ACTIVATION_TOKENS: KVNamespace;
		ASSETS: Fetcher;
	} & EnvVars;
}
```

.gitignore
abdulqshabbir: LOL 🤣  will remove

src/routes/health-check/index.ts
abdulqshabbir: Ooo I like that idea, yeah the fewer type shennanigans the better

src/types/global.d.ts
abdulqshabbir: thanks sounds good

src/routes/health-check/index.ts
```diff
@@ -0,0 +1,52 @@
...
+			},
+			[StatusCodes.UNPROCESSABLE_CONTENT]: {
+				description: 'Error response',
+				content: { 'application/json': { schema: StatusResponseSchema } }
+			}
+		},
+		middleware: [authMiddleware] as const
+	}),
+	(context: Context<EnvironmentBindings>) => {
+		const { success, error } = checkEnvVars(context);
```
madcampos: **super nitpick:** ‏Here we should pass only `context.env`, if we are not checking anything else in the context the parameter should be as specific as possible. Like, a function should only get as parameters things it will use, nothing more.

src/routes/health-check/validation.ts
```diff
@@ -0,0 +1,39 @@
...
+);
+
+function checkEnvVars(context: Context<EnvironmentBindings>) {
+	// NOTE: please add new environment variables here
+	const env: EnvVars = {
+		BASE_URL: context.env.BASE_URL,
+		SENDER_EMAIL: context.env.SENDER_EMAIL,
+		SENDGRID_API_KEY: context.env.SENDGRID_API_KEY,
+		NODE_ENV: context.env.NODE_ENV
+	};
```
madcampos: Tied to the previous comment: if we pass only `env` to this function, we don't need to reconstruct the env object here.

src/routes/health-check/validation.ts
```diff
@@ -0,0 +1,39 @@
+import type { Context } from 'hono';
+import { z } from 'zod';
+import type { ZodSchema } from 'zod';
+type EnvKeys = keyof EnvVars;
+
+// NOTE: please update the schema of new environment variables here
+const envSchema = z.object(
```
madcampos, line 7: **thought:** I'm wondering if here it makes sense to use either [`passthrough`](https://zod.dev/?id=passthrough) or [`strict`](https://zod.dev/?id=strict)‏ options from zod, so we can also point out what extra/unexpected environment variables are present.

Not an issue, just something to discuss.

src/routes/health-check/validation.ts
abdulqshabbir: Oooh yeah that simplifies the code and means the developer only needs to update the schema thanks 👍

src/routes/health-check/validation.ts
abdulqshabbir: Yeah I did think about that. I'm just not too sure yet what the API or DX would look like. If you feel it could be useful to also maybe have like an "error" and "warning" message returned in the response body instead of just an error message I could work on that. Do you feel the "warnings" would bloat the response body?

madcampos: This is for a dev environment only, so I think it is ok to send a warning with a dump of data.

madcampos (approved):

src/routes/health-check/validation.ts
abdulqshabbir: Will create a separate (hopefully a lot smaller!) PR for this

PR #108, opened 2025-03-30
feat: add logout route

### LINKED ISSUE #104: Implement New auth API for signing out

abdulqshabbir:
## Summary

Created logout route

## Tests
- confirm can logout

## Additional information
Resolves https://github.com/torontojs/communityhub-be/issues/104


src/routes/auth/index.ts
```diff
@@ -138,3 +139,18 @@ authRoutes.post('/sign-in', async (context: Context<EnvironmentBindings>) => {
...
+	if (!sessionToken) {
+		return context.json({ message: 'Invalid or missing token' }, StatusCodes.BAD_REQUEST);
+	}
+	// Delete cookie on the server
+	await context.env.SESSION_TOKENS.delete(sessionToken);
+
+	// Delete cookie on the client
+	context.header('Set-Cookie', `auth_token=deleted; HttpOnly; Secure; SameSite=Strict; Expires=Thu, 01 Jan 1970 00:00:00 GMT; Path=/;`);
+
+	return context.json(StatusCodes.NO_CONTENT);
```
bcha92, line 159: Good starting point, but need more context @madcampos  if using `NO_CONTENT` is appropriate.

abdulqshabbir:
@bcha92 and I worked on this together but we weren't sure how to do the fancy "co-authored" commits

kbventures:
@abdulqshabbir @bcha92 

Nice work! We'll start the code review but there's been up date to this https://github.com/torontojs/communityhub-be/issues/104 you may want to have a look at.

src/routes/auth/index.ts
kbventures: @evert any thoughts on "NO_CONTENT"?

evert: 204 No Content in most cases is the same as `200 OK` but you're opting not to return a body.. but `context.json()` is a red flag here!

madcampos: Why not OK?

evert: I would also centralize where all the parameters are set (HttpOnly, Secure, Path) because you want to make sure that where _real_ cookie is set and where it's deleted has the exact same flags. If you hard code these here there's a chance they get out of sync with the real cookie and they could then be considered 'different cookies'. 

(You can have several active cookies with the same name if they have different paths for example)

madcampos:
@abdulqshabbir
> @bcha92 and I worked on this together but we weren't sure how to do the fancy "co-authored" commits

That is from vscode liveshare... But it is just a text referencing the name and email of the other person I think

src/routes/auth/index.ts
madcampos: There is a helper from hono for dealing with cookies: https://hono.dev/docs/helpers/cookie

I would suggest using it, and wrapping the functionality in utility functions that set/unset all of the correct parameters, as Evert pointed out.

evert: > Why not OK?

I think at least for some REST clients, `204 No Content` indicates: "I don't need to parse the body", so that's a nice small benefit. Sometimes if people return `200 OK` they'll try to call `await response.json()` unconditionally.

https://evertpot.com/http/204-no-content

bcha92: > Why not OK?

We can change it to OK, but will need an appropriate 2nd param to return (usually a return body)

madcampos: @bcha92 let's keep the `204 No Content`.

Thanks @evert for the explanation!

PR #110, opened 2025-03-30
Perf:  Optimize sign-in, activate and sign-up routes

### LINKED ISSUE #95: Optimize Session Management: Access Level & DB Queries
[Original conv](https://github.com/torontojs/vms/pull/90/files#r1994150966)

### Summary:
**Issue:** Unnecessary `access_level` checks during login and inefficient database queries.

### Key Points:
- `access_level` is being checked during login, but all users should be able to log in regardless.
- Database queries need optimization since this code runs on every request.

### Proposed Solution:
- Perform a single DB query to fetch user data (e.g., activation status, access level) and use it for all checks.

abdulqshabbir:
## Summary
- combine login sql queries into one join
- update sign-in route to also return generic error messages (can revert if not ok)
- allow user to log in even if no access_level 

## Tests
- verified sign-in route works as expected

## Additional information
- On further inspection I couldn't find a way to optimize the `/activate` route since there is 1 read (`checkEmail`) and 1 write (`activateProfile`)
- Also, I couldn't find a way to further optimize the `/sign-up` route since there is 1 read (`checkEmail`) and 1 batched update (`insertProfile`) 

Resolves https://github.com/torontojs/communityhub-be/issues/95

### DIFF
--- a/src/routes/auth/data.ts
+++ b/src/routes/auth/data.ts
@@ -4,7 +4,10 @@ import { DBTables } from '../../constants/db.ts';
 export async function getLoginInfo(database: D1Database, email: string) {
 	const results = await database
 		.prepare(`
-       SELECT a.password, a.access_level, p.id
+       SELECT
+           a.password as storedPassword,
+           a.access_level as accessLevel,
+           p.id as profileId
        FROM access a
        INNER JOIN
        profile p
@@ -13,7 +16,7 @@ export async function getLoginInfo(database: D1Database, email: string) {
        LIMIT 1
     `)
 		.bind(email)
-		.first<{ password: string, access_level: Access, id: string }>();
+		.first<{ storedPassword: string, accessLevel: Access, profileId: string }>();
 
 	return results;
 }
--- a/src/routes/auth/data.ts
+++ b/src/routes/auth/data.ts
@@ -11,7 +11,7 @@ export async function getLoginInfo(database: D1Database, email: string) {
        FROM access
        INNER JOIN
        profile
-       on profile.id = access.id
+       ON profile.id = access.id
        WHERE profile.email = ? AND profile.activatedAt IS NOT NULL AND profile.deactivatedAt IS NULL
        LIMIT 1
     `)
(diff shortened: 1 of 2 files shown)


src/routes/auth/index.ts
```diff
@@ -102,24 +102,26 @@ authRoutes.post('/sign-in', async (context: Context<EnvironmentBindings>) => {
...
-	const { id: profileId, password: storedPassword } = await getProfileIdPassword(context.env.database, parsedBody.email);
-	if (!profileId || !storedPassword) {
-		return context.json<StatusResponse>({ message: 'Invalid email' }, StatusCodes.UNAUTHORIZED);
-	}
+	const results = await getLoginInfo(context.env.database, parsedBody.email);
 
-	const isProfileValid = await checkProfile(context.env.database, profileId);
-	if (!isProfileValid) {
-		return context.json<StatusResponse>({ message: 'Invalid profile id or Account not activated' }, StatusCodes.UNAUTHORIZED);
+	const genericSignInError = 'Either your email/password combination is invalid, or your account is not active';
```
kbventures, line 107: I'm not a huge fan of these error messages that encompasses several possible error scenario but in this case I cant think of a way around.

src/routes/auth/data.ts
```diff
@@ -1,43 +1,24 @@
...
-        `)
+       SELECT
+           a.password as storedPassword,
+           a.access_level as accessLevel,
+           p.id as profileId
+       FROM access a
+       INNER JOIN
+       profile p
+       on p.id = a.id
+       WHERE p.email = ? and p.activatedAt IS NOT NULL and p.deactivatedAt IS NULL
```
madcampos: **nitpick:** Can we use the table name here instead of single letter references? Sorry to be annoying with that, but I think it would be better for readability.‏

abdulqshabbir: Yep sure a bad habit I've learned somewhere. Agreed it would be better for readability 👍🏼

madcampos (approved):

PR #112, opened 2025-03-30
feat: update Profile schema

### LINKED ISSUE #107: Update Profile Schema
Since [FE#26](https://github.com/torontojs/communityhub-fe/issues/26) and [FE#35](https://github.com/torontojs/communityhub-fe/pull/35) is to be implemented. We should update the object schemas for Profile in preparation of updating existing create/update BE endpoints.

bcha92:
A starting point for updating the Profile Schema in BE conjunction with FE sign up page (see [FE#26](https://github.com/torontojs/communityhub-fe/issues/26) and [FE#35](https://github.com/torontojs/communityhub-fe/pull/35))

### DIFF
--- a/src/routes/profile/validation.ts
+++ b/src/routes/profile/validation.ts
@@ -11,13 +11,45 @@ export const ProfileSchema = BaseDbEntitySchema.merge(z.object({
 		.trim()
 		.min(1, 'Name should be 1 or more characters long.')
 		.describe('The name this person would like to be refered to.'),
+	slack: z
+		.string()
+		.trim()
+		.min(1, 'Slack handle should be 1 or more characters long.')
+		.describe('The slack handle used by this person'),
+	pronouns: z
+		.string()
+		.optional()
+		.describe('Pronouns the user prescribes to'),
+	// Discussion: Change Birthday to ISO Date Object ?
+	birthday: z
+		.string()
+		.describe('Birthday of user'),
+	isBasedOnGTA: z
+		.number()
+		.describe('User is based in GTA'),
+	canJoinLocalEvents: z
+		.number()
+		.describe("User can participate in TorontoJS's local events"),
+	// Discussion: Optional Avatar Upload?
+	avatar: z
+		.string()
+		.optional()
+		.describe('File Name of user avatar'),
+	avatarSource: z
+		.string()
+		.optional()
+		.describe('URL or local upload link of user avatar'),
 	description: z
 		.string()
 		.optional()
 		.describe('A description for this person, may be written in markdown.'),
+	// Discussion: Should LinkedIn, Github, and Site/Portfolio be mandatory links?
 	links: z.array(z.string().url('Invalid url.'))
 		.optional()
-		.describe('A list of links for social media and platforms the person want to make available on the VMS.')
+		.describe('A list of links for social media and platforms the person want to make available on the VMS.'),
+	skills: z.array(z.string())
+		.optional()
+		.describe('A list of skills for this user')
 }));
 
 export type Profile = z.infer<typeof ProfileSchema>;
--- a/src/routes/profile/validation.ts
+++ b/src/routes/profile/validation.ts
@@ -16,9 +16,9 @@ export const ProfileSchema = BaseDbEntitySchema.merge(z.object({
 		.optional()
 		.describe('Pronouns the user prescribes to'),
 	birthday: z
-		.optional(
-			z.string().refine((mmdd) => /([0][1-9]|[1][0-2])-([0][1-9]|[1-2][0-9]|[3][0-1])/u.exec(mmdd), { message: 'Birthday must be valid MM/DD format or nullish value' })
-		)
+		.string()
+		.refine((mmdd) => /([0][1-9]|[1][0-2])-([0][1-9]|[1-2][0-9]|[3][0-1])/u.exec(mmdd), { message: 'Birthday must be valid MM-DD format' })
+		.optional()
 		.describe('Birthday of user'),
 	isBasedOnGTA: z
 		.number()
--- a/src/routes/profile/validation.ts
+++ b/src/routes/profile/validation.ts
@@ -17,7 +17,7 @@ export const ProfileSchema = BaseDbEntitySchema.merge(z.object({
 		.describe('Pronouns the user prescribes to'),
 	birthday: z
 		.string()
-		.refine((mmdd) => /([0][1-9]|[1][0-2])-([0][1-9]|[1-2][0-9]|[3][0-1])/u.exec(mmdd), { message: 'Birthday must be valid MM-DD format' })
+		.refine((mmdd) => Array.isArray(/([0][1-9]|[1][0-2])-([0][1-9]|[1-2][0-9]|[3][0-1])/u.exec(mmdd)), { message: 'Birthday must be valid MM-DD format' })
 		.optional()
 		.describe('Birthday of user'),
 	isBasedOnGTA: z
--- a/src/routes/profile/validation.ts
+++ b/src/routes/profile/validation.ts
@@ -11,16 +11,10 @@ export const ProfileSchema = BaseDbEntitySchema.merge(z.object({
 		.trim()
 		.min(1, 'Name should be 1 or more characters long.')
 		.describe('The name this person would like to be refered to.'),
-	slack: z
-		.string()
-		.trim()
-		.min(1, 'Slack handle should be 1 or more characters long.')
-		.describe('The slack handle used by this person'),
 	pronouns: z
 		.string()
 		.optional()
 		.describe('Pronouns the user prescribes to'),
-	// Discussion: Change Birthday to ISO Date Object ?
 	birthday: z
 		.string()
 		.describe('Birthday of user'),
@@ -34,22 +28,14 @@ export const ProfileSchema = BaseDbEntitySchema.merge(z.object({
 	avatar: z
 		.string()
 		.optional()
-		.describe('File Name of user avatar'),
-	avatarSource: z
-		.string()
-		.optional()
-		.describe('URL or local upload link of user avatar'),
+		.describe('URL or address of user avatar'),
 	description: z
 		.string()
 		.optional()
 		.describe('A description for this person, may be written in markdown.'),
-	// Discussion: Should LinkedIn, Github, and Site/Portfolio be mandatory links?
 	links: z.array(z.string().url('Invalid url.'))
 		.optional()
-		.describe('A list of links for social media and platforms the person want to make available on the VMS.'),
-	skills: z.array(z.string())
-		.optional()
-		.describe('A list of skills for this user')
+		.describe('A list of links for social media and platforms the person want to make available on the Community Hub.')
 }));
 
 export type Profile = z.infer<typeof ProfileSchema>;
--- a/src/routes/profile/validation.ts
+++ b/src/routes/profile/validation.ts
@@ -16,7 +16,9 @@ export const ProfileSchema = BaseDbEntitySchema.merge(z.object({
 		.optional()
 		.describe('Pronouns the user prescribes to'),
 	birthday: z
-		.string()
+		.optional(
+			z.string().refine((mmdd) => /([0][1-9]|[1][0-2])\/([0][1-9]|[1-2][0-9]|[3][0-1])/u.exec(mmdd), { message: 'Birthday must be valid MM/DD format or nullish value' })
+		)
 		.describe('Birthday of user'),
 	isBasedOnGTA: z
 		.number()
--- a/src/routes/profile/validation.ts
+++ b/src/routes/profile/validation.ts
@@ -17,8 +17,11 @@ export const ProfileSchema = BaseDbEntitySchema.merge(z.object({
 		.describe('Pronouns the user prescribes to'),
 	birthday: z
 		.string()
-		.refine((mmdd) => Array.isArray(/([0][1-9]|[1][0-2])-([0][1-9]|[1-2][0-9]|[3][0-1])/u.exec(mmdd)), { message: 'Birthday must be valid MM-DD format' })
 		.optional()
+		.refine(
+			(data) => data ? /^\d{2}-\d{2}$/iu.test(data) : true,
+			{ message: 'Birthday must be valid MM-DD format' }
+		)
 		.describe('Birthday of user'),
 	isBasedOnGTA: z
 		.number()
--- a/src/routes/profile/validation.ts
+++ b/src/routes/profile/validation.ts
@@ -17,7 +17,7 @@ export const ProfileSchema = BaseDbEntitySchema.merge(z.object({
 		.describe('Pronouns the user prescribes to'),
 	birthday: z
 		.optional(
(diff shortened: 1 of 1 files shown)


src/routes/profile/validation.ts
```diff
@@ -11,13 +11,45 @@ export const ProfileSchema = BaseDbEntitySchema.merge(z.object({
 		.trim()
 		.min(1, 'Name should be 1 or more characters long.')
 		.describe('The name this person would like to be refered to.'),
+	slack: z
```
madcampos: **thoughts:** ‏although slack is an important item, I think we should keep it as part of the links list.

@kbventures thoughts?

src/routes/profile/validation.ts
```diff
@@ -11,13 +11,45 @@ export const ProfileSchema = BaseDbEntitySchema.merge(z.object({
...
+		.describe("User can participate in TorontoJS's local events"),
+	// Discussion: Optional Avatar Upload?
+	avatar: z
+		.string()
+		.optional()
+		.describe('File Name of user avatar'),
+	avatarSource: z
+		.string()
+		.optional()
+		.describe('URL or local upload link of user avatar'),
```
madcampos: I was discussing avatar upload with @kbventures, we will only have `avatar` here, and it will be a url.

The two options then will be:
- A user provides their own url for an avatar
- We have an image upload endpoint for this

I'll work on the image upload.

src/routes/profile/validation.ts
```diff
@@ -11,13 +11,45 @@ export const ProfileSchema = BaseDbEntitySchema.merge(z.object({
...
 	description: z
 		.string()
 		.optional()
 		.describe('A description for this person, may be written in markdown.'),
+	// Discussion: Should LinkedIn, Github, and Site/Portfolio be mandatory links?
 	links: z.array(z.string().url('Invalid url.'))
 		.optional()
-		.describe('A list of links for social media and platforms the person want to make available on the VMS.')
+		.describe('A list of links for social media and platforms the person want to make available on the VMS.'),
```
madcampos: Links are not mandatory, and they are an open list.

We do have to save it in a different table though. See how the migration for user is done.

I do think that when we return the user's profile, we should return the data transformed into a nice array of objects.

src/routes/profile/validation.ts
```diff
@@ -11,13 +11,45 @@ export const ProfileSchema = BaseDbEntitySchema.merge(z.object({
...
+	skills: z.array(z.string())
+		.optional()
+		.describe('A list of skills for this user')
```
madcampos: Same as links, skills will be another table on the db.

bcha92: will remove skills as it will be on another table

src/routes/profile/validation.ts
bcha92: will remove `Discussion TOD`.

Considering certain link items like slack handleName are required property on finishing sign up on FE, this may be a topic worth exploring on how to validate (on the BE?) if the user entered a handle name.

src/routes/profile/validation.ts
bcha92: I figured that might be the case, though I was referencing the structure in `\migrations\0002_create-profile.sql`.
Will remove `avatarSource` property

src/routes/profile/validation.ts
bcha92: this one is specific to slack "handle name". As it was listed in the FE as "required" property, and links are optional. Will remove `slack` on next commit.

madcampos: Yeah, I think it may be a soft requirement, or at least we don't have to have it required on the backend.

But my thought here was to save the slack handle along with a list of links so we don't have more specialized fields on the database. Like, the db schema would be sorta generic still but very flexible.

src/routes/profile/validation.ts
madcampos: Yeah, I was thinking about it when reviewing the migration... We can remove it from there as well

src/routes/profile/validation.ts
madcampos: Even if slack handle is required, I think the different tables and all still hold.

Like, on the database, the data is normalized in multiple tables, but when sending a request to the profile endpoints it gets transformed to/into a single object

src/routes/profile/validation.ts
kbventures: @madcampos agreed!

src/routes/profile/validation.ts
kbventures: @bcha92 change VMS to Community Hub.  VMS is more of an internal thing now.

kbventures: @bcha92 @madcampos 

Although slack handle is important I think we shoudl rethink this requirement.  We've seen what happens when something like sendgrid turns pay or go away type scenario.

madcampos: As far as I know, there is no way to save a link to a user on slack...

And right now we don't have any extra infomration about links. It is just the plain url. I think we may want to add some extra metadata, like which platform it comes from, if it is a full url or a handle, etc.

That could solve the problem of saving slack handles in a more manageable way, but also make it generic enough that we can use it for twitter handles, or something along those lines for example.

As for relying on slack: yeah, no. We better not. It takes ownership of the data from us

src/routes/profile/validation.ts
```diff
@@ -11,13 +11,31 @@ export const ProfileSchema = BaseDbEntitySchema.merge(z.object({
 		.trim()
 		.min(1, 'Name should be 1 or more characters long.')
 		.describe('The name this person would like to be refered to.'),
+	pronouns: z
+		.string()
+		.optional()
+		.describe('Pronouns the user prescribes to'),
+	birthday: z
```
abdulqshabbir: idea: If birthday follows a particular format, can use `refine` from zod with a regex pattern to only accept certain kinds of inputs from the frontend.

src/routes/profile/validation.ts
```diff
@@ -11,13 +11,31 @@ export const ProfileSchema = BaseDbEntitySchema.merge(z.object({
...
+		.string()
+		.describe('Birthday of user'),
+	isBasedOnGTA: z
+		.number()
```
abdulqshabbir: idea: If `isBasedOnGTA` is a boolean in sqlite (0 or 1) maybe we can also only allow the number 0 or the number 1

src/routes/profile/validation.ts
```diff
@@ -11,13 +11,31 @@ export const ProfileSchema = BaseDbEntitySchema.merge(z.object({
...
+		.describe('User is based in GTA'),
+	canJoinLocalEvents: z
+		.number()
```
abdulqshabbir: similar comment to `isBasedOnGTA`

src/routes/profile/validation.ts
```diff
@@ -11,13 +11,31 @@ export const ProfileSchema = BaseDbEntitySchema.merge(z.object({
...
+		.describe("User can participate in TorontoJS's local events"),
+	// Discussion: Optional Avatar Upload?
+	avatar: z
+		.string()
```
abdulqshabbir, line 25: suggestion: can add extra validation to conform to URL structure

src/routes/profile/validation.ts
bcha92: definitely something to explore, but currently they're using 0 or 1.

src/routes/profile/validation.ts
bcha92: This is something to discuss with the FE team imo, but as it currently stands, it seems to be a stringifyed ISO Date format

src/routes/profile/validation.ts
bcha92: leaving it as a string here for now... jury's still out as to if this should be a URL (i.e. Gravatar) or a local upload path (i.e. `C://User/Pictures/...`)... what do you think @madcampos @kbventures ?

src/routes/profile/validation.ts
abdulqshabbir: I assumed it was the format in the UI diagram but if not no need to rush it into it

src/routes/profile/validation.ts
abdulqshabbir: I just mean that the database is expecting a 1 (true) or a 0 (false) here. So if you don't limit the numbers you accept to those two (with error handling) and front-end gives you another number (3), internally the database will fail with a SQLite check constraint error and the front-end will get a generic "something went wrong" since we don't send those SQLite errors from BE to FE

src/routes/profile/validation.ts
abdulqshabbir: Yeah I think it might be good to agree on what the shape of the data is that we are receiving/storing. SQLite will store pretty much any text you throw at it for this field and we want to be selective with what text we store in the DB

src/routes/profile/validation.ts
abdulqshabbir: I remember them saying they only want to store something like DD-MM. But maybe @kbventures and @madcampos can chime in

src/routes/profile/validation.ts
abdulqshabbir: > was discussing avatar upload with @kbventures, we will only have avatar here, and it will be a url.

Here's the comment from @madcampos from 2 days ago that I was going off of

abdulqshabbir: I think if it's an upload you won't receive a local path on the client's computer. As I understand it, either you'll receive a URL (like for s3 image upload for example, the front end will directly do the upload to s3 servers using a pre-signed post and give you a URL that holds the file on s3 servers) or you're receive a file through multi part form data.

So depending on what you receive the validation can support that.

bcha92: got it. will change to URL in next commit.

abdulqshabbir: maybe wait for @madcampos I'm not 100% sure either 😆

abdulqshabbir: It's best you ask @madcampos since he's building that service as we speak (image upload) so he has the most up-to-date knowledge and so a comment from 2 days ago could easily be outdated by now

But yes as of 2 days ago it was a URL that you'll receive

src/routes/profile/validation.ts
bcha92: This is assuming in the slack chat we are to follow the `MM/DD` format and not `MM-DD`. But this can easily be swappable in regex.

My only other concern is if calling zod optional first then chaining string with refine is a valid method to bypass potential nullish checks?

madcampos: We have to have a null check in the refine function, something like:
```typescript
refine((data) => data ? /^\d{2}-\d{2}$/iu.test(data) : true)
```

Also, I would like us to follow this format, so when the Temporal API is finally supported, we can easily migrate to it.

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Temporal/PlainMonthDay

src/routes/profile/validation.ts
```diff
@@ -11,13 +11,33 @@ export const ProfileSchema = BaseDbEntitySchema.merge(z.object({
...
 		.min(1, 'Name should be 1 or more characters long.')
 		.describe('The name this person would like to be refered to.'),
+	pronouns: z
+		.string()
+		.optional()
+		.describe('Pronouns the user prescribes to'),
+	birthday: z
+		.string()
+		.refine((mmdd) => Array.isArray(/([0][1-9]|[1][0-2])-([0][1-9]|[1-2][0-9]|[3][0-1])/u.exec(mmdd)), { message: 'Birthday must be valid MM-DD format' })
+		.optional()
```
madcampos: **nitpick:** ‏This follows the same pattern as other schemas adding optional before refine.

Also, the `Array.isArray` check is a hack, as the results of a regular expression are an _special array like_ object[^1]. This behavior works for ✨ compatibility reasons ✨, but **_should not_** be relied upon.

For simple tests if something matches or not, it is better to use [`RegExp.test()`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/RegExp/test).

See this: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/match#description:~:text=If%20you%20need%20to%20know%20if%20a%20string%20matches%20a%20regular%20expression%20RegExp%2C%20use%20RegExp.prototype.test().

```suggestion
		.optional()
		.refine(
			(data) => data ? /^\d{2}-\d{2}$/iu.test(data) : true,
			{ message: 'Birthday must be valid MM-DD format' }
		)
```

[^1]: Technically, it is an instance of `Array`, but it also works as an object with extra properties like `groups`.

madcampos (commented):
Please check this branch, it is incomplete yet and has only partial work for this, but there are some things already done there: https://github.com/torontojs/communityhub-be/tree/feat/missing-profile-fields

src/routes/profile/validation.ts
madcampos: Even if SQLite does not support boolean types natively, we should cast those to boolean and only accept boolean values.

src/routes/profile/validation.ts
madcampos: Here it will _only_ be a fully resolved URL. So the database doesn't have to worry about what it is and the frontend can just throw it in an `img` tag.

Avatar upload will be a separate endpoint that will receive a file and then update the profile setting this property.

> I think if it's an upload you won't receive a local path on the client's computer. As I understand it, either you'll receive a URL (like for s3 image upload for example, the front end will directly do the upload to s3 servers using a pre-signed post and give you a URL that holds the file on s3 servers) or you're receive a file through multi part form data.
>
> So depending on what you receive the validation can support that.

Sorta, kinda, but not exactly. Avatar upload will go to R1 (Cloudflare's S3 compatible storage solution) but will do that _by proxy through the backend_ and not directly. That way we don't have to expose keys, tokens, or anything. The server will handle all the permissions, and can run extra checks if needed.

src/routes/profile/validation.ts
bcha92: changed to null check function

src/routes/profile/validation.ts
bcha92: changing both `isBasedOnGTA` and `canJoinLocalEvents` to boolean.
Will we need a `.default()` value be required (i.e. `false`)?

src/routes/profile/validation.ts
bcha92: If avatar upload is on a separate endpoint, I can proceed to remove this property.

bcha92:
imported `src/routes/profile/validation.ts` from branch "feat/missing-profile-fields"

omitted `skills` property as it was previously discussed this would be combined with `links`

src/routes/profile/validation.ts
bcha92: updated from new commit. from `feat/missing-profile-fields` branch.

abdulqshabbir (approved):

madcampos:
@bcha92
> omitted skills property as it was previously discussed this would be combined with links

Sorry, I think there is some miscommunication here. Both `skills` and `links` will be part of `profile`. But they follow a similar pattern, where the data is stored in a separate table.

PR #113, opened 2025-03-31
Auth check

### LINKED ISSUE #103: Auth api auth check
We need a simple get api in the auth route in order to check if user is authenticated.  Ensuring cookie is still valid. 

For example:

When a user sign-in he is redirected to the home page which has a useEffect making a API call to check if the user is authenticated. If user is not authenticated then the user will be redirected to the sign-in page.

kbventures:
## Summary

Heartbeat route which checks if user is authenticated and returns name, access and avatar information. 

Example flow:

User succesfullly signs in and is redirected to the home page.
The home page makes an api call to /auth/heartbeat before rendering UI based on access or resolve authorized flow.

Github issue: 
https://github.com/torontojs/communityhub-be/issues/103


src/types/data/heartBeat.ts
```diff
@@ -0,0 +1,7 @@
+import type { Access } from './access.ts';
+
+export interface Heartbeat {
```
madcampos: **suggestion:** We could reference the profile schema, as it provides at least `name` and `avatar`‏

src/routes/auth/index.ts
```diff
@@ -138,3 +140,41 @@ authRoutes.post('/sign-in', async (context: Context<EnvironmentBindings>) => {
 
 	return context.json(sessionToken);
 });
+
+authRoutes.get('/heartbeat', async (context: Context<EnvironmentBindings>) => {
```
madcampos: **issue:** Please use openapi as the other endpoint, or else we will have to refactor this endpoint and double the work.‏

src/routes/auth/index.ts
```diff
@@ -138,3 +140,41 @@ authRoutes.post('/sign-in', async (context: Context<EnvironmentBindings>) => {
...
+	const sessionData = await context.env.SESSION_TOKENS.get<SessionData>(sessionToken, 'json') || undefined;
+
+	if (!sessionData) {
+		return context.json({ message: 'Invalid session' }, StatusCodes.UNAUTHORIZED);
+	}
+
+	if (new Date(sessionData.expiry) < new Date()) {
+		await context.env.SESSION_TOKENS.delete(sessionToken);
+		return context.json({ message: 'Session expired' }, StatusCodes.UNAUTHORIZED);
+	}
```
madcampos: **question:** Isn't this code kinda duplicated? We already have this exact same check in a middleware for authorization, no? Can't we reuse that here?‏

src/routes/auth/index.ts
```diff
@@ -138,3 +140,41 @@ authRoutes.post('/sign-in', async (context: Context<EnvironmentBindings>) => {
...
+	const profile = await getProfileById(context.env.database, sessionData.id);
+
+	if (!profile) {
+		return context.json({ message: 'Internal error getting profile that shoudl exist' }, StatusCodes.INTERNAL_SERVER_ERROR);
+	}
+	const {access} = sessionData;
+	const name = profile.name;
+
+	// STUB: Avatar upload and resource under construction
+	const avatar = 'https://gravatar.com/avatar/f8eb6ba9cc4ad24f3b79897a8596ee90?s=400&d=robohash&r=x';
```
madcampos: **suggestion:** What about we add a "default avatar" (can be Toronto JS logo or something) to the assets folder and reference that?‏

bcha92: 💡 idea: could we possibly store a minimized/compressed version of the TorontoJS logo as a "default" local file to cover potential server-side errors/timeouts?
For example if Gravatar (not saying we should use this now or in the future) experiences server/fetch issues that prevent displaying a user avatar, we could use the local file as a "default" backup for avatar placeholders.

src/routes/auth/index.ts
```diff
@@ -138,3 +140,41 @@ authRoutes.post('/sign-in', async (context: Context<EnvironmentBindings>) => {
...
+
+	if (new Date(sessionData.expiry) < new Date()) {
+		await context.env.SESSION_TOKENS.delete(sessionToken);
+		return context.json({ message: 'Session expired' }, StatusCodes.UNAUTHORIZED);
+	}
+
+	const profile = await getProfileById(context.env.database, sessionData.id);
+
+	if (!profile) {
+		return context.json({ message: 'Internal error getting profile that shoudl exist' }, StatusCodes.INTERNAL_SERVER_ERROR);
```
abdulqshabbir: Idea: I think instead of an "Internal error" which usually is interpreted as the server encountered an "unexpected" error we are anticipating this case (by making a check for it). So maybe we can use "Profile not found" with a status code of `NOT_FOUND`

Also, typo with `shoudl`

src/routes/auth/index.ts
```diff
@@ -138,3 +140,41 @@ authRoutes.post('/sign-in', async (context: Context<EnvironmentBindings>) => {
...
+	}
+	const {access} = sessionData;
+	const name = profile.name;
```
abdulqshabbir: nitpick: you're de-structuring on line 173 but not on line 174. Personally I perfer de-structuring when possible but I think consistency is more important. 

(Couldn't find anything else so I was getting desperate... 😂)

src/types/data/heartBeat.ts
kbventures: Avatar doesn't exist yet in this history flow but I've made references to the Name property from Profile validation until we can.

src/routes/auth/index.ts
kbventures: The entire auth page is currently not using the open api pattern. I'll convert the whole route to bring it up to standart.

src/routes/auth/index.ts
madcampos: Yes, and that is the perfect use case for a [service worker](https://developer.mozilla.org/en-US/docs/Web/API/Service_Worker_API)!

Here are some resources for service worker patterns:
- https://serviceworke.rs/
- https://developer.chrome.com/docs/workbox/

src/routes/auth/index.ts
madcampos: image

src/types/data/heartBeat.ts
madcampos, line 1: **nitpick:** the file name casing, it should be `heartbeat.ts` 🙃‏

src/routes/profile/validation.ts
```diff
@@ -37,3 +37,7 @@ export const UpdateProfileSchema = CreateProfileSchema
 	);
 
 export type UpdateProfileData = z.infer<typeof UpdateProfileSchema>;
+
+export const NameSchema = ProfileSchema.pick({ name: true });
+
+export type Name = z.infer<typeof NameSchema>;
```
madcampos: No need to create a new type just for a property of profile. Reference the main profile type instead and then the sub property.

src/types/data/heartBeat.ts
```diff
@@ -0,0 +1,8 @@
+import type { Name } from '../../routes/profile/validation.ts';
```
madcampos:
```suggestion
import type { Profile } from '../../routes/profile/validation.ts';
```

src/types/data/heartBeat.ts
```diff
@@ -0,0 +1,8 @@
...
+import type { Access } from './access.ts';
+
+export interface Heartbeat {
+	name: Name;
```
madcampos:
```suggestion
	name: Profile['name'];
```

Note: It is either bracket, or dot notation for referencing a type's property.

src/types/data/heartBeat.ts
madcampos, line 1: **question:** Why do we have an interface declared here, a schema in the `utils` folder and both are used on the `auth` routes?‏

Shouldn't them be all in the `auth` folder following the same pattern as the other routes?

src/routes/auth/index.ts
```diff
@@ -1,140 +1,286 @@
...
-});
+		try {
+			const body = await context.req.json();
+			parsedBody = SignInSchema.parse(body);
+		} catch (error) {
+			if (error instanceof SyntaxError) {
+				return context.json({ message: `Invalid JSON format: ${error.message}` } satisfies StatusResponse, StatusCodes.BAD_REQUEST);
+			}
+			throw error;
+		}
```
madcampos: Here we don't need to manually parse the body and validate it against a schema. That is done by the openapi for us.

The validated and parsed values are available as returns from the `context.req.valid('json')` function call.

src/routes/auth/index.ts
```diff
@@ -1,140 +1,286 @@
...
-	}
+		try {
+			const body = await context.req.json();
+			parsedBody = CreateProfileSchema.parse(body);
+		} catch (error) {
+			if (error instanceof SyntaxError) {
+				return context.json({ message: `Invalid JSON format: ${error.message}` } satisfies StatusResponse, StatusCodes.BAD_REQUEST);
+			}
+			throw error;
+		}
```
madcampos: Same as previous comment: parsing and validation can be done by openapi

src/routes/auth/index.ts
```diff
@@ -1,140 +1,286 @@
...
+);
+publicAuthRoutes.openapi(
+	createRoute({
+		method: 'get',
+		path: '/',
+		operationId: 'activate',
+		description: 'Received activation email and clicked on activation link',
+		tags: ['Activate'],
+		request: {
+			body: { content: { 'application/json': { schema: SignUpSchema } }, required: true }
```
madcampos:
```suggestion
			params: { content: { 'application/json': { schema: SignUpSchema } }, required: true }
```

We expect a query parameter here, not a request body.

src/routes/auth/index.ts
```diff
@@ -1,140 +1,286 @@
...
+				content: { 'application/json': { schema: StatusResponseSchema } }
+			},
+			[StatusCodes.OKAY]: {
+				description: 'Account activated successfully',
+				content: { 'application/json': { schema: StatusResponseSchema } }
+			}
+		}
+	}),
+	async (context) => {
+		const token = context.req.query('token');
```
madcampos:
```suggestion
		const { token } = context.req.valid('params');
```

src/routes/auth/index.ts
```diff
@@ -1,140 +1,286 @@
...
+		};
+		const sessionData = JSON.stringify(sessionDataObject);
+		await context.env.SESSION_TOKENS.put(sessionToken, sessionData);
 
-	const isValid = await validatePassword(parsedBody.password, storedPassword);
-	if (!isValid) {
-		return context.json<StatusResponse>({ message: 'Invalid email or password' }, StatusCodes.UNAUTHORIZED);
+		context.header('Set-Cookie', `auth_token=${sessionToken}; HttpOnly; Secure; SameSite=Strict; Expires=${tokenExpiryISO}; Path=/;`);
+
+		return context.json({ message: 'Sign in successful' } satisfies StatusResponse, StatusCodes.CREATED);
```
madcampos: I don't remember in which PR exactly, but there is a comment to change this specific call to use the hono cookie helper.

src/routes/auth/index.ts
```diff
@@ -1,140 +1,286 @@
...
+		};
+		await sgMail.send(msg);
+
+		return context.json({ message: 'Created a new profile and sent an email for confirmation' } satisfies StatusResponse, StatusCodes.OKAY);
 	}
+);
+publicAuthRoutes.openapi(
+	createRoute({
+		method: 'get',
+		path: '/',
```
madcampos:
```suggestion
		path: '/activate',
```

The way it is right now that route would be: `GET /auth/`

src/routes/auth/index.ts
```diff
@@ -1,140 +1,286 @@
...
-	return context.json(sessionToken);
+// Protected auth routes
+export const protectedAuthRoutes = new OpenAPIHono<EnvironmentBindings>({
+	defaultHook: statusResponseFormatter
 });
+
+protectedAuthRoutes.openapi(
+	createRoute({
+		method: 'get',
+		path: '/',
```
madcampos:
```suggestion
		path: '/heartbeat',
```

The way it is right now, that route would be: `GET /auth/`

src/routes/auth/index.ts
```diff
@@ -1,140 +1,286 @@
...
+		const sessionData = context.get('session');
+
+		if (!sessionData) {
+			return context.json({ message: 'Invalid session' } satisfies StatusResponse, StatusCodes.UNAUTHORIZED);
+		}
+
+		const profile = await getProfileById(context.env.database, sessionData.id);
+
+		if (!profile) {
+			return context.json({ message: 'Internal error getting profile that shoudl exist' } satisfies StatusResponse, StatusCodes.INTERNAL_SERVER_ERROR);
```
madcampos: **question:** Ain't we‏ already checking for the profile existence in the authorization middleware?

src/routes/auth/index.ts
```diff
@@ -1,140 +1,286 @@
...
+		}
+		const { name } = profile;
+		const { access } = sessionData;
```
madcampos, line 282: **suggestion:** What about we combine the work with this PR?

https://github.com/torontojs/communityhub-be/pull/110‏

That way we make a single query to the server instead of multiple ones for the same-ish information

src/routes/auth/validate.ts
```diff
@@ -22,4 +22,16 @@ export const SignInSchema = z.object({
...
+		.min(1, 'Email must be at least one character long')
+		.email('Invalid Email'),
+	password: z
+		.string()
+		.min(1, 'Password must be at least one character long')
+});
+
 export type SignInData = z.infer<typeof SignInSchema>;
+
+export type SignUpData = z.infer<typeof SignUpSchema>;
```
madcampos: **nitpick:** Please refactor to be schema followed by type. That way we keep the schema and type closer to one another.‏

src/routes/auth/index.ts
kbventures: It is indeed.  New iteration has that resolved.

src/routes/auth/index.ts
kbventures: Just did that! :)

src/index.ts
```diff
@@ -54,16 +55,19 @@ app.doc('/open-api.json', {
...
 // Public routes
-app.route('/auth', authRoutes);
+app.route('/auth', publicAuthRoutes);
 app.route('/roles', publicRoleRoutes);
 app.route('/profiles', publicProfileRoutes);
 app.route('/teams', publicTeamRoutes);
 
-// Private routes
+// Protected routes (after auth middleware)
+app.use('/*', authMiddleware);
```
abdulqshabbir: suggestion: we can remove this line and instead run the `authMiddleware` inside of the `createRoute` function of any protected route

src/routes/auth/index.ts
```diff
@@ -1,142 +1,288 @@
...
 import { addHours } from 'date-fns';
-import { type Context, Hono } from 'hono';
+import { Hono } from 'hono';
 import { generateEmailHtml } from '../../email-templates/confirm-email.ts';
+import { authorizeVolunteer } from '../../middleware/createMiddleware.ts';
 import type { SessionData } from '../../types/data/session';
 import { hashPassword, validatePassword } from '../../utils/password-hashing.ts';
-import { StatusCodes, type StatusResponse } from '../../utils/responses.ts';
-import { insertProfile } from '../profile/data.ts';
+import { type HeartbeatResponse, HeartbeatResponseSchema, StatusCodes, type StatusResponse, statusResponseFormatter, StatusResponseSchema } from '../../utils/responses.ts';
```
abdulqshabbir: question: the `HeartbeatResponse`, `HeartbeatResponseSchema` aren't generic response utils. Should they be in a different file?

src/routes/auth/index.ts
```diff
@@ -1,142 +1,288 @@
...
-	try {
-		const body = await context.req.json();
-		parsedBody = CreateProfileSchema.parse(body);
-	} catch (error) {
-		if (error instanceof SyntaxError) {
-			return context.json<StatusResponse>({ message: `Invalid JSON format: ${error.message}` }, StatusCodes.BAD_REQUEST);
+publicAuthRoutes.openapi(
+	createRoute({
+		method: 'post',
+		path: '/',
```
abdulqshabbir: issue: sign-up route is no longer pointing to `/auth/sign-up`

src/routes/auth/index.ts
```diff
@@ -1,142 +1,288 @@
...
+		return context.json({ message: 'Account activated successfully' } satisfies StatusResponse, StatusCodes.OKAY);
 	}
+);
 
-	// Remove token after successful activation
-	await context.env.ACTIVATION_TOKENS.delete(token);
+publicAuthRoutes.openapi(
+	createRoute({
+		method: 'post',
+		path: '/',
```
abdulqshabbir: issue: I think Marco already mentioned this but this doesn't point to `/auth/sign-in` anymore

src/routes/auth/index.ts
```diff
@@ -1,142 +1,288 @@
...
+			[StatusCodes.UNAUTHORIZED]: {
+				description: 'No cookies found, invalid or missing token, invalid session or session expired. ',
+				content: { 'application/json': { schema: StatusResponseSchema } }
+			},
+			[StatusCodes.OKAY]: {
+				description: 'Heartbeat succesful. User authenticted and name, avatar and access returned to Client.',
+				content: { 'application/json': { schema: HeartbeatResponseSchema } }
+			}
+		},
+		middleware: [authorizeVolunteer] as const
```
abdulqshabbir: suggestion: add an `authMiddleware` check before the `authorizeVolunteer`

src/routes/auth/index.ts
```diff
@@ -1,142 +1,288 @@
...
+		if (!profile) {
+			return context.json({ message: 'Internal error getting profile that should exist' } satisfies StatusResponse, StatusCodes.NOT_FOUND);
+		}
+		const { name } = profile;
+		const { access } = sessionData;
+
+		// STUB: Avatar upload and resource under construction
+		const avatar = 'https://gravatar.com/avatar/f8eb6ba9cc4ad24f3b79897a8596ee90?s=400&d=robohash&r=x';
+
+		return context.json({ message: 'Heartbeat succesful', access, name, avatar } satisfies HeartbeatResponse, StatusCodes.OKAY);
```
abdulqshabbir: question: does the front-end need a `Heartbeat successful` message? Or is it enough to return `access`, `name`, `avatar`?

abdulqshabbir (changes requested):

src/index.ts
abdulqshabbir: I've done this for the existing routes for any new ones we create we can always add the `authMiddleware` inside the `createRoute`

src/routes/auth/index.ts
abdulqshabbir: Yeah we'll take care of this in the logout PR

src/routes/auth/validate.ts
```diff
@@ -22,4 +22,20 @@ export const SignInSchema = z.object({
...
+	email: z
+		.string({ required_error: 'Email is required' })
+		.min(1, 'Email must be at least one character long')
+		.email('Invalid Email'),
+	password: z
+		.string()
+		.min(1, 'Password must be at least one character long')
+});
+
+export const ActivateSchema = z.object({ id: z.string().uuid('Invalid ID format') });
```
abdulqshabbir:
```suggestion
export const ActivateSchema = z.object({ token: z.string().uuid('Invalid ID format') });
```

abdulqshabbir: before the `/auth/activate` was taking a param `token` as input

src/routes/auth/index.ts
```diff
@@ -1,142 +1,288 @@
...
-			return context.json<StatusResponse>({ message: `Invalid JSON format: ${error.message}` }, StatusCodes.BAD_REQUEST);
+publicAuthRoutes.openapi(
+	createRoute({
+		method: 'post',
+		path: '/',
+		operationId: 'signUp',
+		description: 'Sign up to community hub account',
+		tags: ['Sign-up'],
+		request: {
+			body: { content: { 'application/json': { schema: SignUpSchema } }, required: true }
```
abdulqshabbir, line 33: suggestion: I don't think the `/auth/sign-up` route should only accept email & password. It also needs to take in `name` and `description` for example to create a profile. Instead of using `SignUpSchema` maybe you could use `CreateProfileSchema` instead?

src/routes/auth/index.ts
```diff
@@ -1,142 +1,288 @@
...
+	async (context) => {
+		let parsedBody: CreateProfileRequestBody;
 
-	const emailExists = await checkEmail(context.env.database, parsedBody.email);
-	if (emailExists) {
-		return context.json<StatusResponse>({ message: 'Duplicate email' }, StatusCodes.CONFLICT);
-	}
+		try {
+			const body = await context.req.json();
+			parsedBody = CreateProfileSchema.parse(body);
```
abdulqshabbir, line 55: suggestion: No need to parse the body here. Let Open API handle that based on the schema you pass on line 29 of this file

src/routes/auth/index.ts
```diff
@@ -1,142 +1,288 @@
...
+		}
+	}),
+	async (context) => {
+		let parsedBody: SignInData;
 
-	return context.json({ message: 'Account activated successfully' }, StatusCodes.OKAY);
-});
+		try {
+			const body = await context.req.json();
+			parsedBody = SignInSchema.parse(body);
```
abdulqshabbir, line 191: suggestion: don't parse the request body here let Open API do it based on the schema you declare. The parsed body will be available with something like `const parsedBody = context.req.valid('json')`

src/routes/auth/validate.ts
abdulqshabbir: Before the `auth/activate` route was taking `token` as a param

src/routes/auth/index.ts
```diff
@@ -1,142 +1,288 @@
...
+				content: { 'application/json': { schema: StatusResponseSchema } }
+			},
+			[StatusCodes.OKAY]: {
+				description: 'Account activated successfully',
+				content: { 'application/json': { schema: StatusResponseSchema } }
+			}
+		}
+	}),
+	async (context) => {
+		const { token } = context.req.valid('params');
```
abdulqshabbir:
```suggestion
		const { token } = context.req.valid('param');
```

abdulqshabbir: I think you can access the params with `param` instead (naming might be a little weird here haha)

src/routes/auth/index.ts
kbventures: Agreed.  Removed.

src/routes/auth/index.ts
kbventures: It should have name(is missing), email and password.  Create Profile Schema is a different flow.

kbventures: @abdulqshabbir @madcampos 

The initial sign up flow calls for name, email and password:

image

Are we not going to create another route for the post sign up profile creation ?

src/routes/auth/index.ts
kbventures: @abdulqshabbir 

Just confirming.  You want to replace those two lines with 

```const parsedBody  = await context.valid.json()```

File: src/routes/auth/index.ts (hunk 28 above)
Comment (kbventures) [reply src/routes/auth/index.ts:55]: ```const parsedBody  = await context.valid.json()``` ???

File: src/routes/auth/index.ts (hunk 27 above)
Comment (abdulqshabbir) [reply src/routes/auth/index.ts:33]: Oh oops I think I may be wrong here sorry. I can double check in the morning. Maybe I was thinking create profile route and not sign up route.

File: src/routes/auth/index.ts (hunk 29 above)
Comment (abdulqshabbir) [reply src/routes/auth/index.ts:191]: We can maybe talk a bit tomorrow about it during our coding session?! But basically all the logic around using a Zod schema to parse the request body.

File: src/routes/auth/index.ts (hunk 17 above)
Comment (kbventures) [reply src/routes/auth/index.ts]: @madcampos 

The session data doesn't currently containe the name property

```
export interface SessionData {
	id: string;
	email: string;
	access: Access;
	expiry: ISODate;
}
```

It since we'll likely be utilizing heartbeat check a lot it might make sense to include it in the SessionData object instead of making a data base call?

File: src/routes/auth/index.ts (hunk 4 above)
Comment (kbventures) [reply src/routes/auth/index.ts]: Created https://github.com/torontojs/communityhub-be/issues/118

File: src/routes/auth/index.ts (hunk 27 above)
Comment (abdulqshabbir) [reply src/routes/auth/index.ts:33]: I double-checked the dev branch and for the `/auth/sign-up` route it is using `CreateProfileSchema` to parse the request body. It seems that the profile is also created in the same route (whether we want that or not I'm not sure I'm just listing the existing functionality)

File: src/routes/auth/index.ts (hunk 29 above)
Comment (abdulqshabbir) [reply src/routes/auth/index.ts:191]: similar to Marco's comment the entire try catch block is not needed and can be replaced with something like this:
		`const parsedBody = context.req.valid("json")`

File: src/routes/auth/index.ts [hunk 31]
```diff
@@ -1,142 +1,290 @@
...
+);
+publicAuthRoutes.openapi(
+	createRoute({
+		method: 'get',
+		path: '/activate',
+		operationId: 'activate',
+		description: 'Received activation email and clicked on activation link',
+		tags: ['Activate'],
+		request: {
+			params: ActivateSchema
```
abdulqshabbir, line 108:
```suggestion
			query: ActivateSchema
```

I think the activate route is no longer working -- it should accept query params (this change and one below has to be made)

src/routes/auth/index.ts
```diff
@@ -1,142 +1,290 @@
...
+				content: { 'application/json': { schema: StatusResponseSchema } }
+			},
+			[StatusCodes.OKAY]: {
+				description: 'Account activated successfully',
+				content: { 'application/json': { schema: StatusResponseSchema } }
+			}
+		}
+	}),
+	async (context) => {
+		const { token } = context.req.valid('param');
```
abdulqshabbir, line 134:
```suggestion
		const { token } = context.req.valid('query');
```

To make query params work

src/routes/auth/index.ts
kbventures: I might be lost in semantics here. 

image

src/routes/auth/index.ts
kbventures: Intdocing the suggested code introduces a chain of new errors.

src/routes/auth/index.ts
kbventures: Intdocing the suggested code introduces a chain of new errors.

src/routes/auth/index.ts
madcampos: No. We don't want to change the `SessionData` object.

What makes more sense is to also load the profile, or at least part of it, for authenticated calls and expose that to the route through the authentication middleware.

That is how it is usually done at least.

src/index.ts
madcampos: We should actually add the middlewares in a very explicit way to all routes that need it. The openapi definition allows for a middleware array, we should use that.

src/routes/auth/index.ts
madcampos: The sign-up route has only email and password. It should create a new profile, if one does not exist and silently fail otherwise. (It is an attack vector to try and create a new profile with an existing email just to check if it exists or not).

**Then** there is the complete profile step where more information will be provided.

src/routes/auth/index.ts
madcampos: Read the hono openapi docs for more examples and details on the specific method to use: https://github.com/honojs/middleware/tree/main/packages/zod-openapi

That should help with the errors

src/routes/auth/index.ts
madcampos: That is on me, was a dyslexic moment, sorry.

abdulqshabbir: No worries their naming conventions are a little weird haha

PR #114, opened 2025-04-03
Fix: auth token should not be sent in response body

abdulqshabbir:
## Summary

auth token should not be sent in the response body as it is more likely to be intercepted. The token is already being sent as an http only secure cookie.

### DIFF
--- a/src/routes/auth/index.ts
+++ b/src/routes/auth/index.ts
@@ -136,5 +136,5 @@ authRoutes.post('/sign-in', async (context: Context<EnvironmentBindings>) => {
 
 	context.header('Set-Cookie', `auth_token=${sessionToken}; HttpOnly; Secure; SameSite=Strict; Expires=${tokenExpiryISO}; Path=/;`);
 
-	return context.json(sessionToken);
+	return context.json({ message: 'Successfully signed in' });
 });


src/routes/auth/index.ts
```diff
@@ -136,5 +136,5 @@ authRoutes.post('/sign-in', async (context: Context<EnvironmentBindings>) => {
 
 	context.header('Set-Cookie', `auth_token=${sessionToken}; HttpOnly; Secure; SameSite=Strict; Expires=${tokenExpiryISO}; Path=/;`);
 
-	return context.json(sessionToken);
```
bcha92, line 139: Looks good. But why did this exist in the first place?
Would this cause any issues for the FE team down the line (i.e. persistent cookie)?

abdulqshabbir: the `Set-Cookie` header should handle the setting of the cookie on the client device. The client should still be able to send that cookie to our backend (managed by the browser)for auth purposes.

I'm not too sure why it's being sent in the response body but I don't think it should be there

evert (approved):

src/routes/auth/index.ts
madcampos: No it shouldn't LOL

I think in this case an empty success response is _OKAY_ (pun intended). That will set the cookie for http requests only (meaning it won't be visible on the js side) and will make things work just fine.

madcampos (approved):

src/routes/auth/index.ts
kbventures: @abdulqshabbir I noticed it as well.  It somehow got passed the code reviews. :)

https://github.com/torontojs/communityhub-be/pull/113

kbventures (approved):

PR #115, opened 2025-04-03
Fix: cannot delete profile

abdulqshabbir:
## Summary
Cannot delete profiles using the `/profiles/{{profileId}}` `DELETE ` route.
This is because when the profile table (parent table) is deleted there are still foreign key references to that table that remain from children tables (e.g. access table)

The `ON DELETE CASCADE` will delete children tables when the parent table is deleted. If we want a different action open to changing it but cascade makes sense I think in this context.

## Tests

Steps to reproduce:
1. create an account in vms, and activate
2. set your access level for the account to admin (`npx wrangler d1 execute vms-local --command="update access set access_level = 'admin'"`)
3. sign in using that access level
4. call the delete profile API

Before: could not delete because of foreign key constraints
Now: can delete

If you'd like to check if the fix is working need to switch to this branch, delete db, run migrations and also run steps 1-4


evert (commented):
I don't really have an opinion about this change for this project.

However I never use CASCADE deletes because I usually want every delete to be a deliberate choice.

For a community project wiping all user data seems ethical and a good move, but for commercial projects you sometimes might want to keep analytics. So I always explicitly delete every record in code in a single transaction because it forces me to think about it

kbventures:
@abdulqshabbir @madcampos there's also the question of wether we want delete profile in most cases as opposed to simply making them inactive.  One of the side effects of the data is historical context and it that historical persective of who did what and when is an important pillar of the project.

madcampos:
I think in this case we don't want to cascade updates as the referenced keys on the original tables will not change, they should be unique.

As for deletion, this is a larger discussion in terms of which policy we will have for a couple reasons:
- Simply deleting data breaks other parts of the platform, specifically historical data.
- It is a nice to have option to disable/re-enable profiles instead of wiping them.
- I agree with Evert that we do have to have a policy for erasing data. But that should be a deliberate and very serious choice.

As a potential solution (up for discussion, of course) we would provide easely only a way to deactivate an account. Reactivating or entirely deleting the data, would be done by contacting an organizer.

We can have a script for that admin action ready and tested, for sure, but I think this should be a _partially manual_ process, not something baked into the platform.

As for historical data consistency, we can confirm with the person when deleting their data, if they want some references like their name to still be kept, or if they want a complete redaction of the data.

If they allow for keeping some information we can update their profile to remove every extra piece of information and only keep something like the name, or even a short version of the name. If they want redaction, then we redact everything, including their name and exclude things like pasword.

An aside to that is: the historical data will be exported in the future to json files, those files will be publically available on github, so even if we remove all information from an user, it will still be kept in git. We can of course rewrite git history to redact things, but there is always a chance someone will have a backup of the old data.

kbventures (approved):

madcampos:
Closing this PR as we don't want to _simply delete_ data. Most of the entries are marked with "soft deletes" instead.

PR #117, opened 2025-04-05
Fix/active sessions and token refresh

### LINKED ISSUE #94: Handling Active Sessions and Token Refresh on Re-sign-in
This is a summary from a conversation
:

https://github.com/torontojs/vms/pull/90/files#r1994176163

### Problem:
When a user is already signed in and has an active session, calling the sign-in endpoint again may cause issues:

- **Security Vulnerability:** Old session tokens remain valid, potentially allowing unauthorized access if stolen.
- **Resource Waste:** Unused tokens accumulate until expiration, increasing storage and management overhead.
- **User Experience Concern:** No notification or confirmation is provided when a new session overwrites an existing one, which may confuse users.

### Proposed Solutions:
1. **Check for Active Sessions:** Before issuing a new token, check if an active session exists. If it does, notify the user they are already signed in.
2. **Single Active Session:** Allow only one active session per user, clearing it upon expiration or logout.
3. **Multiple Device Sessions:** Allow users to have unique session tokens per device (e.g., PC/mobile) with expiration after a period of inactivity (e.g., 90 days).
4. **Token Refresh:** If a user signs in again with valid credentials, refresh the existing token instead of creating a new one.

### Open Questions:
- How to reliably verify the user's device (e.g., IP address is unreliable, especially on mobile)?
- Should sessions expire after a fixed duration (e.g., 90 days)?

abdulqshabbir:
## Summary

New changes added:
- Do not allow users with sessions to create new sessions
- Add explicit check/deletions for expired sessions
- Extend the session lifespan from 1 day to 90 days (can change if too long)
- Allow automatic session extension. If user uses API on day 46 (after the half-way point) the session gets extended by another 90 days

Refactors:
- Created auth utils file with `createSession` and `getSession` functions which can be reused in other locations in codebase for consistent auth checks

## Tests
- To test you can change `SESION_LIFESPAN_IN_DAYS` to 10,  all `addDays` to `addSeconds` and all `differenceInDays` to `differenceInSeconds` so that sessions only last for 10 seconds

## Additional information
Resolves https://github.com/torontojs/communityhub-be/issues/94


src/utils/auth.ts
```diff
@@ -0,0 +1,149 @@
...
+	if (isExpired) {
+		throw new Error('Invalid session');
+	}
+
+	const TWO = 2;
+	const HALF_SESSION_LIFESPAN_IN_DAYS = Math.floor(SESION_LIFESPAN_IN_DAYS / TWO);
+
+	const daysUntilTokenExpiry = differenceInDays(new Date(sessionExpiryISO), new Date());
+
+	return daysUntilTokenExpiry < HALF_SESSION_LIFESPAN_IN_DAYS;
```
madcampos: **thought:** ‏I think here we can skip complex checks and just extend the session every time. If we have shorter sessions, like one day, or one week, it is only extended a little bit at a time and not a huge chunk‏

src/utils/auth.ts
```diff
@@ -0,0 +1,149 @@
...
+	// Delete session on client
+	setCookie({
+		context,
+		name: SESSION_COOKIE_NAME,
+		value: 'deleted',
+		expires: new Date(0)
+	});
+}
+
+async function getValidatedSession(context: Context<EnvironmentBindings>) {
```
madcampos: **nitpick:** let's keep names more direct
```suggestion
async function getSession(context: Context<EnvironmentBindings>)‏
```

src/utils/auth.ts
```diff
@@ -0,0 +1,149 @@
...
+	const sessionToken = getCookie(context, SESSION_COOKIE_NAME);
+	if (!sessionToken || sessionToken === 'deleted') {
```
madcampos: **question:** Why do we have a check for `deleted` here? When will a token be set to deleted?‏

src/utils/auth.ts
```diff
@@ -0,0 +1,149 @@
...
+	if (shouldSessionExtend(session.expiry)) {
+		await extendExistingSession({
+			sessionKV: context.env.SESSION_TOKENS,
+			sessionToken,
+			session
+		});
+		return {
+			session,
+			sessionToken
+		};
```
madcampos: **issue:** Don't mutate function parameters.

If you need output data that is a result from transforming an input, then the function should return the transformed data and that should be used instead.

Mutating a variable outside of the local context of a function is very bad as it leads to hard to find and debug errors. You are basically transforming the variable in a black box that anything can happen to it.‏

src/utils/auth.ts
```diff
@@ -0,0 +1,149 @@
...
+	}
+
+	return {
+		session,
+		sessionToken
```
madcampos: **question:** Are we using `sessionToken` anywhere? If not, this could be simplified to return only `session`.‏

src/utils/auth.ts
abdulqshabbir: Sounds good will make the change

src/utils/auth.ts
abdulqshabbir: Honestly, I'm partial to having a blank cookie with no value when we log out but we decided as a group that it would be set to the string 'deleted'. If you'd like me to remove that check it's not a big deal we just have to also update the logout branch.

src/utils/auth.ts
abdulqshabbir: Thanks for catching this. Completely agree and I'll be mindful of not mutating in the future.

src/utils/auth.ts
abdulqshabbir: I think at some point it was needed but no longer after the refactor. Thanks for catching.

src/utils/auth.ts
abdulqshabbir: I'm just concerned about one thing here. Since this function runs in middleware on every request we need two KV store calls (get session and extend session) on every request if we don't have some way of telling when we should extend the session

abdulqshabbir: I'll switch the expiry duration to 24 hours if that sounds good to you

madcampos: > I'm just concerned about one thing here. Since this function runs in middleware on every request we need two KV store calls (get session and extend session) on every request if we don't have some way of telling when we should extend the session

I think that is an okay compromise to have.

src/utils/auth.ts
madcampos: I just don't get it, why would the cookie be blank? When is it set to the string "deleted"?

Like, if the cookie is invalid, it still holds a value and it is fine, but no entry will exist on the KV store. That already tells us that the cookie is invalid.

src/utils/auth.ts
abdulqshabbir: Ok sounds good 👍🏼

src/utils/auth.ts
abdulqshabbir: Ohh I see. Yeah agreed, that check is unnecessary

src/utils/cookie.ts
```diff
@@ -0,0 +1,32 @@
...
+	value: string;
+	expires: Date;
+}
+
+function setCookie({
+	context,
+	name,
+	value,
+	expires
+}: SetCookieInput) {
```
abdulqshabbir, line 24: Oh btw @madcampos, we kind of have this same cookie helper in the logout branch but with a different name and function signature.

Wondering if you had any strong preference as I have to reconcile the two.

In the logout branch the function signature doesn't use an object as input and is a little more consistent with the `setHonoCookie` from hono. The name is different as well, it's called `presetSetCookie`.

```js
function presetSetCookie(
	context: Context<EnvironmentBindings>,
	name: string,
	value: string,
	expires: Date
): void
```

abdulqshabbir: Also, do you think we should add an automatic expiry to the sessions in the KV store? Currently I think sessions will just hang around even if expired to my understanding.

madcampos: > Wondering if you had any strong preference as I have to reconcile the two.

I usually prefer a configuration object when there are 3+ parameters as long function parameters used to be harder to read without [inlay hints](https://code.visualstudio.com/docs/editing/editingevolved#_inlay-hints)[^1]. But here it is okay I guess, I have no strong opinions, so do what you feel looks nicer 🤷‍♂️

> Also, do you think we should add an automatic expiry to the sessions in the KV store? Currently I think sessions will just hang around even if expired to my understanding.

I don't know if there is something like an "auto remove when expired" mechanism for KV, if there is _and_ we can retrieve the expiration date metadata set on the KV entry, then we could simplify the code to a simple check if the key exists or not. Like, if it does not exist, it is either expired or invalid, either way the user have to re-authenticate.

[^1]: In my time we only had notepad++! [cue old man yelling at clouds]

madcampos (approved):

PR #121, opened 2025-04-12
Fix: activate route uses query params

abdulqshabbir:
## Summary

Activate route was taking `token` as a query param before (if we want to keep existing behavior)


madcampos:
I'm refactoring this code, will integrate the changes from this PR.

abdulqshabbir:
Sounds good thanks

PR #123, opened 2025-04-12
Fix: do not allow sign-in if user session exists

abdulqshabbir:
## Summary

- Don't allow users to sign-in if they already have a session
- Redundant KV store call to create session


abdulqshabbir:
closing this as some parts are outdated compared to the grand backend unification!

PR #127, opened 2025-04-12
Feat: Automatically expire session keys

abdulqshabbir:
## Summary

Automatically expire session keys in KV store.  Currently, we manually delete sessions from the KV store if the session is expired _when_ a user makes a request that goes through auth middleware.

Now, the keys will auto-delete from the KV store when the session expires. This will happen automatically even if the user makes no subsequent requests to our api. 

## Tests

In the `getSessionExpirySecondsEpoch` function I changed the expiry to one minute into the future (instead of 1 day) for testing purposes (minimum duration in future for KV store) and it worked ok for that scenario.

## Additional Info

Resolves https://github.com/torontojs/communityhub-be/issues/124


src/utils/auth.ts
```diff
@@ -18,12 +19,16 @@ export interface SessionData {
...
 	token: string;
 }
 
-function isSesionExpired(sessionExpiryISO: string) {
-	return new Date() > new Date(sessionExpiryISO);
+function getSessionExpirySecondsEpoch() {
+	const tokenExpiry = new Date();
+	tokenExpiry.setHours(tokenExpiry.getHours() + SESSION_LIFESPAN_IN_HOURS);
+
+	const tokenExpirySecondsEpoch = Math.floor(tokenExpiry.valueOf() / MILISECONDS_IN_SECOND);
```
madcampos: **suggestion**: Due to ✨ math reasons ✨ I think here we may have a loss of precision because of the division.

So my suggestion here would be to return the result in milliseconds, and then only do the division when we pass the value to KV.

That way we also avoid the second multiplication to get this value back.

src/utils/auth.ts
```diff
@@ -59,29 +64,28 @@ async function extendExistingSession({
...
 	// Update session on server
 	await context.env.SESSION_TOKENS.put(
 		sessionToken,
-		JSON.stringify(updatedSession)
+		JSON.stringify(session),
+		{
+			expiration: expirySecondsSinceEpoch,
+			metadata: {
+				expiration: expirySecondsSinceEpoch
+			}
```
madcampos: **question**: is this property needed? I think KV only checks for the other 
`expiration` property and it ignores `metadata`. Should be safe to remove it.

src/utils/auth.ts
```diff
@@ -136,18 +134,23 @@ export async function createSession({
...
-				expiry: tokenExpiry.toISOString(),
 				token: sessionToken
 			} satisfies SessionData
-		)
+		),
+		{
+			expiration: expirySecondsSinceEpoch,
+			metadata: {
+				expiration: expirySecondsSinceEpoch
+			}
```
madcampos: Same here, `metadata` should be safe to remove.

src/utils/auth.ts
```diff
@@ -59,29 +64,28 @@ async function extendExistingSession({
...
-	};
-
+	const expirySecondsSinceEpoch = getSessionExpirySecondsEpoch();
 	// Update session on server
 	await context.env.SESSION_TOKENS.put(
 		sessionToken,
-		JSON.stringify(updatedSession)
+		JSON.stringify(session),
+		{
+			expiration: expirySecondsSinceEpoch,
```
madcampos: **suggestion**: From the KV docs, it also allows us to set the `expirationTtl`. Which is the number of seconds from now. Wouldn't that be preferable as it can be set to a fixed constant instead, simplifying the logic?

src/utils/auth.ts
abdulqshabbir: Oooh good catch 💡

src/utils/auth.ts
abdulqshabbir: Yeah I agree it would be simpler code-wise. It comes with the tradeoff of not knowing the absolute time for the session expiry. But since this is handled behind the scenes it would be unlikely we need to know so I'm ok with the tradeoff if you're also ok with it 😄

madcampos: We are not handling token expiration nor deletion right? Like, if it does not exist, we asusme it is expired.

If that is the case, and we don't forsee needing to know for some code reason if the token is expired, then I would say to make the change 🤷‍♂️

abdulqshabbir: yep, we're not handling the token deletion when it expires anymore (happens automatically). Changed the session expiry to using relative time which simplifies the code a bit 👍

src/utils/auth.ts
```diff
@@ -2,7 +2,8 @@ import type { Context } from 'hono';
 import { getCookie, setCookie } from 'hono/cookie';
 import type { CookieOptions } from 'hono/utils/cookie';
 
-const SESSION_LIFESPAN_IN_HOURS = 24;
+// eslint-disable-next-line @typescript-eslint/no-magic-numbers
+const SESSION_LIFESPAN_IN_SECONDS_FROM_NOW = 60 * 60 * 24;
```
madcampos:
```suggestion
const SESSION_LIFESPAN_IN_SECONDS = 60 * 60 * 24;
```

**nitpick**: let's just keep "in seconds", the "from now" part would imply that is a date starting now, with added time. This variable is only the time to add.

src/utils/auth.ts
```diff
@@ -28,13 +29,8 @@ export interface SessionData {
...
 
-function getSessionExpirySecondsEpoch() {
-	const tokenExpiry = new Date();
-	tokenExpiry.setHours(tokenExpiry.getHours() + SESSION_LIFESPAN_IN_HOURS);
-
-	const tokenExpirySecondsEpoch = Math.floor(tokenExpiry.valueOf() / MILISECONDS_IN_SECOND);
-
-	return tokenExpirySecondsEpoch;
+function getSessionExpiryAsDate() {
+	return new Date(Date.now() + SESSION_LIFESPAN_IN_SECONDS_FROM_NOW * MILISECONDS_IN_SECOND);
```
madcampos: **question**: Why not use the date methods for adding seconds?

src/routes/auth/index.ts
```diff
@@ -108,7 +108,7 @@ authRoutes.openapi(
 		}
 
 		const userAlreadyActivated = await checkActiveEmail(context.env.Database, email);
-		if (!userAlreadyActivated) {
+		if (userAlreadyActivated) {
 			// INFO: Hide non existing emails to reduce attack surface from guessing registered emails.
```
kbventures, line 112: Do we still need this note?

kbventures (approved):
Aside from fixing suggestions from @madcampos this is good to go.

src/routes/auth/index.ts
madcampos: I left it there, it is good to have to explain why we bail out of the function earlier

PR #128, opened 2025-04-12
Feat: add warnings for env variables that may not be in use

abdulqshabbir:
## Summary
Add warning message for variables that may not be in use for local/dev environment

## Tests
Tested by calling endpoint

## Additional information
Resolves https://github.com/torontojs/communityhub-be/issues/119


src/routes/health-check/responses.ts
```diff
@@ -0,0 +1,10 @@
+import { StatusResponseSchema } from 'src/utils/responses';
+import { z } from 'zod';
+
+export const HealthCheckResponseSchema = StatusResponseSchema.merge(
+	z.object({
+		warning: z.string().optional()
```
bcha92: Shouldn't the warning message be mandatory instead?

If it is optional, are we going to pass a nullish value/empty string if no string is sent for Health Check?

bcha92 (approved):
Except for that one comment (a bit of a curious query on my part), everything else looks good.

src/routes/health-check/validation.ts
```diff
@@ -21,10 +21,30 @@ export const EnvSchema = z.object({
...
 
 export type EnvVars = z.infer<typeof EnvSchema>;
 
 export function checkEnvVars(env: Context<EnvironmentBindings>['env']) {
-	return EnvSchema.safeParse(env);
+	const { success, error, data: parsedEnv } = EnvSchema.safeParse(env);
+
+	if (success) {
+		const allEnvKeys = Object.keys(parsedEnv);
+		const usedEnvKeys = Object.keys(EnvSchema.shape);
```
madcampos:
```suggestion
		const allEnvKeys = Object.keys(parsedEnv);
		const expectedEnvVars = Object.keys(EnvSchema.shape);
```

I think here we should rename the variable to something else, as it is not listing the unused variables, but the variables that are defined in the schema.

src/routes/health-check/responses.ts
```diff
@@ -0,0 +1,10 @@
+import { StatusResponseSchema } from 'src/utils/responses';
+import { z } from 'zod';
+
+export const HealthCheckResponseSchema = StatusResponseSchema.merge(
```
madcampos: **suggestion**: maybe we can add the status response to also include `warning` or other information like a `details` string or object...

Like, it is intended to be used as a general response for the platform for everything that is not a response that returns data (or paginated response).

So, we can return extra information here, or make the status check response a _data_  response.

src/routes/health-check/responses.ts
abdulqshabbir: I guess the warning here is only being returned in local/dev environment when all the "required" environment variables are set but some "optional" ones are not.  So not all responses will have the `warning` field which is why I marked it as optional.

So in the respond body we can choose to omit the `warning` field if we want to and still return a `message` field.

src/routes/health-check/validation.ts
abdulqshabbir: sounds good 👍

src/routes/health-check/responses.ts
abdulqshabbir: Good idea! I've add a `warning` for now and maybe when the need arises we can add a `details` field when we feel the `message` field is insufficient

src/routes/health-check/validation.ts
```diff
@@ -21,10 +21,30 @@ export const EnvSchema = z.object({
...
+			message: `✅ All required environment variables are set`,
+			warning: `🤷 These variables may not be in use: ${unexpectedEnvVars.join(', ')}`
+		};
+	}
+
+	const errorMsg = error.issues.map(({ path, message }) => `${path}: ${message}`).join('. ');
+	return {
+		success: false,
+		message: `❌ Required variables missing: ${errorMsg}`
+	};
```
madcampos:
```suggestion
	return {
		message: `❌ Required variables missing`,
		errors: error.issues.map(({ path, message }) => ({
			path,
			message
		}))
	};
```

There is an `errors` property already on the status response. That and the status code of the message should be enough to tell there is a problem. Also, sending structured data instead of a string is better for parsing and debugging.

src/utils/responses.ts
```diff
@@ -202,6 +202,8 @@ export const StatusCodes = {
 export const StatusResponseSchema = z.object({
 	message: z.string()
 		.describe('A message reporting the status of the operation.'),
+	warning: z.string().optional()
+		.describe('A warning message regarding the status of the operation.'),
```
madcampos:
```suggestion
	warning: z.array(z.record(z.string(), z.string()))
		.optional()
		.describe('A warning message regarding the status of the operation.'),
```

I think here we can do the same as `errors` and provide structured data (i.e.: an object), that way it is easier to parse and handle specific values instead of delaing with string parsing.

src/utils/responses.ts
```diff
@@ -202,6 +202,8 @@ export const StatusCodes = {
...
 	errors: z.record(z.string(), z.array(z.string()))
```
madcampos:
```suggestion
	errors: z.array(z.record(z.string(), z.string()))
```

This was wrong, oops...

src/routes/health-check/validation.ts
```diff
@@ -21,10 +21,30 @@ export const EnvSchema = z.object({
...
+		const expectedEnvVars = Object.keys(EnvSchema.shape);
+
+		const unexpectedEnvVars = allEnvVars
+			.filter((variable) => !expectedEnvVars.includes(variable));
+
+		return {
+			success: true,
+			message: `✅ All required environment variables are set`,
+			warning: `🤷 These variables may not be in use: ${unexpectedEnvVars.join(', ')}`
+		};
```
madcampos:
```suggestion
		return {
			success: true,
			message: `✅ All required environment variables are set`,
			warning: unexpectedEnvVars.map((envVar) => ({
				message: '🤷 This variables may not be in use',
				variable: envVar
			}))
		};
```

madcampos:
So, the comments relate to the ideas behind status message:

1. Using structured data for errors and other messages, we can better parse and handle them specifically.
   It also helps with tools that consume logs/responses like Datadog, New Relic,  and friends.
3. Multiple error can happen at once, so they should be an array of objects that gives more information about each specific error.
   For example, a stack trace can be represented as a list of errors/items.
4. The structured data is _not defined_ as it depends on the endpoint returning it, but that is fine, as figuring out things based on one object is easier than parsing strings.
   It is really not that important to be fixed in a single format, as long as we have some common building blocks to base off of we should be fine.
5. An _error response_ should be identifiable by the response status code and as a fallback, the presence of the `errors` property.
   Sending a `success: false` property with a `200 - OK` status code is morally wrong and a sin in the eyes of our lord and saviour Linus Torvalds.
   _**Anyone caught doing that should be forced to read the lunix mailing list as panitence.**_

Side note: I may have made mistakes on the review suggestions as I did it through github ui editing markdown, so keep that in mind 🙈

abdulqshabbir:
> 1. Using structured data for errors and other messages, we can better parse and handle them specifically. It also helps with tools that consume logs/responses like Datadog, New Relic,  and friends.

I've seen this pattern repeatedly of an errors array with a semi-structured format but never took time to understand the benefits until now, thanks!

> 2. Multiple error can happen at once, so they should be an array of objects that gives more information about each specific error.  For example, a stack trace can be represented as a list of errors/items.

👍 

> 3. The structured data is _not defined_ as it depends on the endpoint returning it, but that is fine, as figuring out things based on one object is easier than parsing strings.  It is really not that important to be fixed in a single format, as long as we have some common building blocks to base off of we should be fine.

Yeah, I think I understand now that transforming from object (even semi-structured) =>  string (or whatever client wants) is 100x easier than transforming from string => object (or what client wants).

>    Sending a `success: false` property with a `200 - OK` status code is morally wrong and a sin in the eyes of our lord and saviour Linus Torvalds.

LOL agreed that the status code coveys the success/failed status of the network request

>   _**Anyone caught doing that should be forced to read the lunix mailing list as panitence.**_

Don't know if this would be painful, insightful or both 😆

madcampos:
> Don't know if this would be painful, insightful or both 😆

Both, but in the worst way possible LOL

madcampos (approved):

PR #129, opened 2025-04-12
Feature: first event log implementation

madcampos:
## Summary

Add event log entries for joining Toronto JS, creating a team and deleting a team.

## Tests

No, tests are missing.


madcampos:
For context, here is an updated drawing of what the tables and relationships should look like:

null

migrations/0004_create-role.sql
```diff
@@ -1,5 +1,6 @@
 -- Migration number: 0004 	 2025-01-30T00:58:39.397Z
 
+DROP INDEX IF EXISTS idx_role_profile_and_team ON role;
```
abdulqshabbir: suggestion: I think we can change/remove the `ON role` part of the drop index query. It seems to be causing issues with migration. I'm not 100% sure if sqlite has a concept of dropping indexes scoped to a table.

There is a `schema-name` part of the drop index statement but I'm not too sure how to use it. Ref: https://sqlite.org/lang_dropindex.html

Similar comment applies to other migration files that are using the `drop index` statement with an `on [table]`

src/routes/event-log/data.ts
```diff
@@ -0,0 +1,78 @@
...
+			object: teamId,
+			objectSource: LogItemSource.TEAM
+		});
+	}
+
+	static closeTeam(database: D1Database, profileId: string, teamId: string) {
+		return EventLog.createLogEntry(database, {
+			subject: profileId,
+			subjectSource: LogItemSource.PROFILE,
+			verb: 'created',
```
abdulqshabbir, line 63: question: should the verb be closed?

migrations/0004_create-role.sql
```diff
@@ -13,9 +14,9 @@ CREATE TABLE IF NOT EXISTS role (
 	-- The role description
 	description TEXT,
 	-- The UUID of the team this role belongs to
-	teamId TEXT NOT NULL COLLATE BINARY,
+	team_id TEXT NOT NULL COLLATE BINARY,
```
abdulqshabbir: suggestion: in the seed file I think there is still an old reference to `teamId` which can be changed to `team_id`. There is also an older reference to `profileId` which can be changed to `profile_id`

src/routes/event-log/data.ts
```diff
@@ -0,0 +1,78 @@
+import { DBTables, DEFAULT_TEAM_ID, generateBaseDBfields } from '../../constants/db.ts';
+import { type EventLog as EventLogType, LogItemSource } from './validation.ts';
+
+export class EventLog {
+	static createLogEntry(
```
abdulqshabbir, line 5: Praise: I like the `createLogEntry` abstraction and also the static methods on the class in general :)

src/routes/profile/data.ts
```diff
@@ -42,7 +43,7 @@ export async function insertProfile(database: D1Database, { email, name, passwor
 				?, ?, ?, ?, ?
 			)
 		`).bind(id, schemaVersion, 'volunteer', password, email),
-
+		EventLog.joinTorontoJS(database, id),
 		database.prepare(`
 			INSERT INTO ${DBTables.ROLE} (
 				id, schemaVersion, happenedAt, insertedAt,
```
abdulqshabbir, line 63: suggestion: the insert call in the line below is still inserting into `teamId` column which can be changed into `team_id` and `profileId` which can be changed to `profile_id`.

src/routes/team/data.ts
```diff
@@ -1,25 +1,37 @@
...
-		)
-	`)
-		.bind(id, schemaVersion, happenedAt, insertedAt, name, description)
-		.run();
+	const results = await database.batch([
+		database.prepare(`
+			INSERT INTO ${DBTables.TEAM} (
+				id, schemaVersion, happenedAt, insertedAt,
+				name
+				${description ? ', descrition' : ''}
```
abdulqshabbir: suggestion: typo on `descrition`

src/routes/profile/data.ts
abdulqshabbir: There is also this query `const existingData` that is filtering by `profileId`.

src/routes/event-log/data.ts
```diff
@@ -0,0 +1,78 @@
...
+				object, objectSource
+			)
+			VALUES (
+				?, ?, ?, ?,
+				?, ?,
+				?,
+				?, ?
+			)
+		`)
+			.bind(id, schemaVersion, happenedAt, insertedAt, subject, subjectSource, verb, object, objectSource);
```
abdulqshabbir: nitpick: the bind parameters can be formatted to match how they are inserted for ease of readability: 4-2-1-2

src/routes/team/index.ts
```diff
@@ -130,14 +130,15 @@ teamRoutes.openapi(
 	}),
 	async (context) => {
 		const { id } = context.req.valid('param');
+		const { id: profileId } = context.get('session');
```
abdulqshabbir: nitpick: can use the `getSession` function from auth utils.

src/routes/team/index.ts
```diff
@@ -171,8 +172,9 @@ teamRoutes.openapi(
 		middleware: [authMiddleware, authorizeAdmin] as const
 	}),
 	async (context) => {
+		const { id: profileId } = context.get('session');
```
abdulqshabbir: nitpick: can use the `getSession` function from auth utils.

migrations/0004_create-role.sql
madcampos: ~~We can do it after table creation, I think that will not break the indexes.~~

~~Like, I'm not sure if dropping a table also removes the indexes, if it does, we can definitely remove the drop indexes, if not, we have to find a way to also take care of that.~~

Confirmed by SQLite docs, dropping a table removes it's indexes.

src/routes/event-log/data.ts
madcampos: Yes, that was a typo

migrations/0004_create-role.sql
madcampos: ~~I'm actually thinking about changing everything to be camelCase instead of snake_case...~~

Edit: changed everything to canelCase

src/routes/profile/data.ts
madcampos: Yeah, I'm updating everything to be camelCase to avoid inconsistencies and feel more "javascript-y"

src/routes/event-log/data.ts
madcampos: I think dprint is not formatting this propperly, it either put everything in one line or breaks it in separate lines... Have to double check though

migrations/0004_create-role.sql
abdulqshabbir: ooh nice didn't know that but makes sense that dropping a table would also remove indexes 👍

migrations/0004_create-role.sql
abdulqshabbir: sounds good 👍

src/routes/event-log/data.ts
abdulqshabbir: yeah if dprint is getting it in the way it's not worth the pain 🤣

src/routes/event-log/validation.ts
```diff
@@ -0,0 +1,33 @@
+import { z } from 'zod';
+import { IdAndSchemaVersionSchema, InsertionTimestampsSchema } from '../../utils/db.ts';
+
+export const LogItemSource = {
```
kbventures, line 4: I'm assuming "special" has to do with possibly making inserts in history prior to the inception of the community hub?(Founds, past directors, etc)

kbventures (dismissed):

src/routes/event-log/validation.ts
madcampos: I was thinking "special" to be more like the reference to Toronto JS itself, it is not a team, but one of the few "fixed" things with a fixed id that never changes.

It may apply to other cases as well, or we can remove the "special" category.

Issue #111, opened 2025-03-30
SendGrid is no longer free


kbventures:
We need a new solution. 

We're considering this solution:

https://resend.com/

We can move to an open source solution in the future.

kbventures:
https://developers.cloudflare.com/workers/tutorials/send-emails-with-resend/
https://dev.to/pierre/gmail-with-cloudflare-email-routing-4413
https://developers.cloudflare.com/email-routing/

PR #133, opened 2025-04-23
Feature: grand backend unification

madcampos:
## Summary

Hear ye, hear ye, tis _Le Grand Backèndé Unífìcationne_

Okay, jokes aside. Here is a big refactor of the backend code to integrate multiple branches and pieces of work into a single cohesive part.

Here is a list of changes:
- Dependencies have been bumped to the latest versions.
- Enabled [Eraseable Syntax flag](https://www.typescriptlang.org/tsconfig/#erasableSyntaxOnly).
- Moved `constants/db.ts` to `utils/db.ts`.
- All routes now follow the same OpenAPI pattern
- Add Swagger UI to the `/docs` endpoint
- Healthcheck is now the root endpoint for better ergonomics. We want the root of the API to run health tests and confirm everything is okay.
- Set `.nvmrc` to always point to the lts version of node. This fixes problems with the node version used on Cloudflare and adds compatibility for those using nvm instead of volta.
- Add indexes to fields that are potentially used in lots of queries to improve db performance.
- Implement queries for profile that handle skills and links and save them to their respective tables.
- Add checks for queries to profiles so we always filter closed and not activated profiles out.
- Remove sendgrid dependency and modularize email sending code. The email function is now just a dummy promise that will always resolve and log the authorization token to the console. This can make testing works even without an email provider.
- Refactor compound word db fields to be normalized as `camelCase` instead of `snake_case` so it feels more "js". Table names are still `snake_case`.
- Combine activation and authorization code.
- Remove all `types` files except for `globals.d.ts`. Aside from globals, which sets the context variables, we want all types to be closer to the code they refer to.
- Rename `roles` to `team_members` as it connects `profiles` and `teams`, so it makes more sense to give it a name that reflects the relationship instead of the generic `roles`. (Please, prove me wrong here and suggest a better name 😅)
- Merge PRs #126 and #128 
- Update checks for existing profiles and teams to be less generic. Before there was a `checkEntityExistence` test, now we have `doesProfileExist` and `doesTeamExist`.
- Update cloudflare config to make the app deployable.
- Update all context bindings to be `PascalCase` instead of `UPPERCASE`. The reasoning is that they are objects with methods and resemble more classes than environment variables, so we use casing here to distinguish between environment variables (`UPPERCASE`) and objects (`PascalCase`).
- Add a basic test mock for environment bindings. This can be used to generate mocks for testing routes that require database calls, assets, kv, or other things that bind to `context`.
- Rework unit test file into a stub to be updated later. It has the needed building blocks for adding more unit tests in the future.

## Tests

- No unit tests
- No integration tests
- Basic requests are working on the app (i.e. healthcheck works)
- The app deploys correctly to cloudflare 🥳

## Additional information

Feel free to ask questions and bash the code 😅


src/email/index.ts
```diff
@@ -0,0 +1,38 @@
...
+	senderEmail: _senderEmail
+}: AccountConfirmationEmailParams) {
+	console.log({
+		email,
+		token
+	});
+
+	return Promise.resolve(undefined);
+
+	// TODO: update email service
```
bcha92, line 23: Until we can get a suitable substitute, I think we're fine leaving this out?

bcha92, line 23: Until we can get a reliable substitute to Sendgrid, I think we're fine leaving this out as a marker?

Issue Link: [BE-#111](https://github.com/torontojs/communityhub-be/issues/111)

Any update on resend.com implementation @kbventures ?

bcha92 (dismissed):
Other than that one comment about Sendgrid, it looks good.

src/email/index.ts
madcampos: Pretty much, yes. The comment is here as a reminder.

As for functionality, this function logs to the console the key for activation, so we would be able to, in the worst case, just copy and paste it to activate an account.

kbventures: @bcha92 we're going to try setting up resend Thursday. We thought there were issues on there side but it turns out we had email issues on our side. Setting up a mail service top of list priority.

src/routes/auth/index.test.ts
```diff
@@ -0,0 +1,22 @@
+import { describe, expect, test } from 'vitest';
```
kbventures, line 1: I'm not sure how you were able to push this code I get this error locally:

```
Parsing error: Unknown compiler option 'erasableSyntaxOnly'.eslint

```

src/middleware/auth.ts
```diff
@@ -1,16 +1,11 @@
 import type { Context, Next } from 'hono';
```
kbventures, line 1: Getting this error again locally

```
Parsing error: Unknown compiler option 'erasableSyntaxOnly'.eslint
```

src/routes/auth/index.ts
```diff
@@ -131,125 +97,103 @@ publicAuthRoutes.openapi(
...
 
-		const activated = await activateProfile(context.env.database, email);
+		const activated = await activateProfile(context.env.Database, email);
 		if (!activated) {
 			return context.json({ message: 'Failed to activate account' }, StatusCodes.INTERNAL_SERVER_ERROR);
 		}
 
 		// Remove token after successful activation
-		await context.env.ACTIVATION_TOKENS.delete(token);
+		await context.env.ActivationTokens.delete(token);
```
kbventures, line 122: Should we consider adding a expirationTTL attribute to the activation token  store entry instead of introducing checks / delete?

https://developers.cloudflare.com/kv/api/write-key-value-pairs/

Search: expiration

kbventures:
What is the rational for ?

"Enabled [Eraseable Syntax flag](https://www.typescriptlang.org/tsconfig/#erasableSyntaxOnly)."

src/routes/auth/index.test.ts
madcampos: Sorry, I've updated all of the packages/dependencies, so to run it okay reinstall them:

```shell
npm install --save-dev
```

madcampos:
@kbventures 

> What is the rational for ?
> 
> "Enabled [Eraseable Syntax flag](https://www.typescriptlang.org/tsconfig/#erasableSyntaxOnly)."

The rationale is that node is starting to support typescript natively and building tools that also support typescript usually support a sub-set of features. Specifically: they support type annotations and that's about it.

Tooling seems to be converging to this, where they do support typescript but only things that can be removed easily and do not affect the language. For us that means we remove `enum`s as we are not using any of [the other things](https://devblogs.microsoft.com/typescript/announcing-typescript-5-8/#the---erasablesyntaxonly-option) that are not erasable[^1].

So, for us it doesn't change the code a lot and makes the syntax more future proof, by moving the codebase closer to standard js.

[^1]: Aside from `enum`s, the others are not very common at all, unless in some old ts codebases. Like, I've _never_ seen `namepsace`s with code nor `import` aliases before, _ever_. Class property parameter I've seen before, but they are considered kind of a bad practice.

src/middleware/auth.ts
madcampos: Same as [here](https://github.com/torontojs/communityhub-be/pull/133#discussion_r2057357392): reinstall the packages

src/routes/auth/index.test.ts
kbventures: I had done that already actually but the error was still there.  I'm noticing it's no longer there in the morning. Sometimes vs code doesn't update right away.

src/routes/profile/data.ts
```diff
@@ -41,84 +67,182 @@ export async function insertProfile(database: D1Database, { email, name, passwor
...
+
+	if (!profile) {
+		return undefined;
+	}
 
-	return results?.[0];
+	profile.links = (results[1]?.results as ProfileLink[] | undefined ?? []).map(({ url }) => url);
+	profile.skills = (results[2]?.results as ProfileSkill[] | undefined ?? []).map(({ skill }) => skill);
+
+	return transformProfile(profile);
```
lylaminju, line 189: We can use LEFT JOIN in a single query.
It works well, but I'm not sure if this kind of typing is a best practice.

```ts
const result = await database.prepare(`
    SELECT
	    p.*,
	    GROUP_CONCAT(DISTINCT l.url) as links,
	    GROUP_CONCAT(DISTINCT s.skill) as skills
    FROM ${DBTables.PROFILE} p
	    LEFT JOIN ${DBTables.PROFILE_LINKS} l ON p.id = l.profileId
	    LEFT JOIN ${DBTables.PROFILE_SKILLS} s ON p.id = s.profileId
    WHERE
	    p.id = ?
	    AND p.activatedAt IS NOT NULL
	    AND p.deletedAt IS NULL
    GROUP BY p.id
  `)
  .bind(id)
  .first();

if (!result) {
  return undefined;
}

const links = (result['links'] as string | null)?.split(',');
const skills = (result['skills'] as string | null)?.split(',');

return transformProfile({
  ...(result as Profile),
  links,
  skills
});
```

src/routes/profile/data.ts
```diff
@@ -41,84 +67,182 @@ export async function insertProfile(database: D1Database, { email, name, passwor
...
 }
 
 export async function getAllProfiles(database: D1Database) {
-	const { results } = await database.prepare(`SELECT * FROM ${DBTables.PROFILE}`).run<Profile>();
+	const results = await database.batch([
```
lylaminju, line 193: We can use LEFT JOIN in a single query as well
I tested the `/profiles` api with this query, and it worked well! But, I'm not sure if this kind of typing is a best practice.

```ts
const { results } = await database.prepare(`
    SELECT
	    p.*,
	    GROUP_CONCAT(DISTINCT l.url) as links,
	    GROUP_CONCAT(DISTINCT s.skill) as skills
    FROM ${DBTables.PROFILE} p
	    LEFT JOIN ${DBTables.PROFILE_LINKS} l ON p.id = l.profileId
	    LEFT JOIN ${DBTables.PROFILE_SKILLS} s ON p.id = s.profileId
    WHERE
	    p.activatedAt IS NOT NULL
	    AND p.deletedAt IS NULL
    GROUP BY p.id
  `)
  .run();
  
return results.map((profile) => {
    const links = (profile['links'] as string | null)?.split(',');
    const skills = (profile['skills'] as string | null)?.split(',');
    
    return transformProfile({
	    ...(profile as Profile),
	    links,
	    skills
    });
});
```

src/routes/profile/data.ts
madcampos: Interesting, yes, we can definitely use a join here!

When I tried it, the results were actually annoying. Like, it returned multiple rows for the same profile... But I only tried an `INNER JOIN`, so maybe `LEFT JOIN`s perform better here.

Another option is to use the [`json_array`](https://www2.sqlite.org/matrix/json1.html#the_json_array_function) function from sqlite... I couldn't get it to work though 🥲

But yeah, my sql skills are not great, so I'll give this a try, thanks!

src/routes/profile/data.ts
madcampos: @lylaminju what do you mean by "not sure if this kind of typing is a best practice"?

lylaminju: @madcampos Oh, sorry for not explaining it clearly. I meant "this kind of typing" as some things related to Typescript skills, like type assertion `profile['links'] as string | null` or putting a new object into `transformProfile()` with type assertion `...(profile as Profile)`

I had to use those type assertions because the type of `results` is `Record<string, unknown>[]` and the `profile` type is `Record<string, unknown>`

kbventures (approved):

src/routes/profile/data.ts
madcampos: I like the `JOIN`, but I'm hitting a similar problem as the one I had at first when I tested it: if we don't add the `DISTINCT` part to `GROUP_CONCAT` it will return a list of skills * links length, where every item will show up multiple times.

My concern here is for users with a large number of skills or links how much will that affect performance?

See the screenshot below:
image

madcampos: Another point is we cannot set the separator when using `DISTINCT`, it has to always be a "`,`"... I'm concerned about misintepreting the data when splitting it, so I'd prefer if there is an option that leave less room for parsing/errors

src/routes/profile/data.ts
madcampos: Ah, okay, got it.

There is a way to make it work that is very simple: add an [Intersection Type](https://www.typescriptlang.org/docs/handbook/2/objects.html#intersection-types) to the `first` call. Like this:

```typescript
first<Profile & { skills?: string, links?: string }>();
```

The type here will assign it to the results of calling the SQL query, it will be everything from the profile AND the extra fields, `skills` and `links`.

src/routes/auth/index.ts
madcampos: For activation? No, as we want to actually activate the profile.

Maybe we should also add a limit so a profile can't be in the "not activated" state indefinitely.

PR #150, opened 2025-05-11
feat: add get profile/self api endpoint

### LINKED ISSUE #149: Add an API endpoint to get logged-in user's profile
Add the following API endpoint to get logged-in user's profile

GET /profile/self

manvinderjit:
Add get api endpoint to fetch profile data of the currently logged-in user.

## Summary

Add a GET api endpoint to fetch profile of the currently logged-in user

## Tests

Pending.

## Additional information

uses route `/profiles/self`

### DIFF
--- a/src/routes/profile/index.ts
+++ b/src/routes/profile/index.ts
@@ -64,6 +64,42 @@ profileRoutes.openapi(
 profileRoutes.openapi(
 	createRoute({
 		method: 'get',
+		path: '/self',
+		operationId: 'Get profile data',
+		summary: 'Get profile data of logged-in user',
+		description: 'Retrieves the profile data of the currently logged-in user',
+		tags: ['Profile'],
+		responses: {
+			[StatusCodes.NOT_FOUND]: {
+				description: 'Internal error getting profile that should exist',
+				content: { 'application/json': { schema: StatusResponseSchema } }
+			},
+			[StatusCodes.UNAUTHORIZED]: {
+				description: 'No cookies found, invalid or missing token, invalid session or session expired. ',
+				content: { 'application/json': { schema: StatusResponseSchema } }
+			},
+			[StatusCodes.OKAY]: {
+				description: 'Successful response',
+				content: { 'application/json': { schema: generateDataResponeSchema(ProfileSchema) } }
+			}
+		},
+		middleware: [authMiddleware] as const
+	}),
+	async (context) => {
+		const sessionData = getSession(context);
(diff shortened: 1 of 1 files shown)


bcha92:
While the code looks okay in general, my biggest concern is DRY, `profile/self` and `profile/{id}` endpoints are duplicated with slight variations in adding unauthorized status code. Instead of repeating, I suggest modifying the existing `profile/{id}` endpoint if all you are doing is fetching the profile and if the user is "unauthorized", they only see the information the user wants the public to see, and if they are "authorized", the user sees all information.

Now, if this was about editing your own profile, I'd focus on a separate endpoint using `put` or `patch` (I'm not sure which ones you've ended up discussing with the team with) to modify the profile.

I mean, that's my take on this in a nutshell, but I'll also hear out from others and see if this is the direction we want to take since I could be wrong about this in hindsight.

bcha92 (commented):
While the code looks okay in general, my biggest concern is DRY, since we already have a `profile/{id}` endpoint that does exactly the same thing as this endpoint, except for slight variations in adding unauthorized status code when fetching user info from `getProfileById()` and using the sessionData to validate the user is viewing one's own profile instead of the profile ID params.

In my opinion, it might be easier to modify the existing `profile/{id}` endpoint and use SessionData as a validation for other things, like editing one's own profile. If the sessionData does not match the user they are viewing, they should only see the public profile.

I mean, that's my take on this in a nutshell, but I'll also hear out from others and see if this is the direction we want to take since I could be wrong about this if the intended direction was unclear to me, and if there are any privacy-specific concerns I haven't addressed with my proposal.

madcampos (approved):

madcampos:
@bcha92 I think for a DRYer code here, we could extract the logic from the handling function and use it in both places.

We still want to have a specific route for getting the user's own profile as the `{id}` parameter is validated as a UUID and adding conditional validation may make the code messier overall with extra ifs.

But to give more context on why this endpoint is suggested: as the user session is kept in an http only cookie, the client app has no visibility on the current logged in user, or even if there is a logged in user.

To solve this problem, we have a `heartbeat` endpoint (needs a better naming though) but that endpoint only returns limited information about the user, like the avatar, name, email, and access level. It is useful to figure out the logged in user, but incomplete for the profile.

This created another situation: how do we get the complete profile for the current user?
To solve that, the idea came to use a new `profiles/self` endpoint, where it returns the full profile for the current logged in user.

PR #151, opened 2025-05-12
Feature: setup email configuration for local testing

### LINKED ISSUE #145: Local email sending for tests
We want to add a way to unblock local (and CI) test from depending on email sending service.

One idea is to stub emails or add flags that can be configured to send/not send the email so we can bypass this dependency.

madcampos:
## Summary

Add an env var for testing emails locally, with that set, emails will only be logged to the console and not sent.

## Tests

Manually tested.

## Additional information

- Cleaned up env vars
- Updated typescript config
- Added auto type generation from wrangler command
- Added script to run install and type generation on pull
- Infer server url instead or hard coding it

Closes: #145


lylaminju (approved):
LGTM! The idea to auto generate `env.d.ts` file is clever. It ensures that the type definitions are always up-to-date with the current branch's configuration.

kbventures (approved):

madcampos:
@lylaminju that was actually from the cloudflare docs, not everything is awful with them LOL

PR #154, opened 2025-05-13
unit tests for sign in/sign up endpoints

### LINKED ISSUE #152: Unit test - authentication

bcha92:
## Summary

Unit tests for auth route, primary focus on sign-in, sign-up, and sign-out routes

## Tests

Sign-up route (todo)
- Failure: registration should be valid email
- Failure: registration should fail if email is already registered
- Failure: registration should have valid password
- Failure: registration should have name
- Registration successful
- failure: activation should not be successful if incorrect token
- failure: activation should not be successful if token expired
- Activation successful

Sign-in route (todo)
- Success
- Failure: Should not log in with incorrect username and/or password
- Failure: Should not log in if account is not activated
- Success: Should log in successfully with correct username and password

Sign-out route (todo)
- Bad request if invalid token provided
- User is successfully logged out

Password Hashing Utilities
- hashPassword should return a string in the format "salt:hash"
- hashPassword should generate different hashes for the same password
- validatePassword should return true for correct password
- validatePassword should return false for incorrect password
- validatePassword should throw error for invalid hash format

Auth Data Functions
- getLoginInfo
  - returns login info for activated, non-deleted user
  - returns null if no matching user found
- getHeartbeatInfo
  - returns heartbeat info for activated user
  - returns null if user is not found
- checkExistingEmail
  - returns true if email exists
  - returns false if email does not exist
- checkActiveEmail
  - returns false if email is inactive
- activateProfile (todo)
  - returns true when update succeeds
  - returns false when update fails

## Additional information

Added `npm test` script to run `vitest`


src/routes/auth/index.test.ts
```diff
@@ -5,16 +5,81 @@ import { MockEnvBindings } from '../../utils/testing.ts';
 
 const MOCK_ENV = new MockEnvBindings();
 
+describe('Sign-up route', () => {
+	const signUpResponse = async (email: string, password: string, name: string) => {
+		const response = await app.request('/auth/sign-up', {
```
madcampos, line 10: I think tests using `app.request` were not working. Are they now?

bcha92: `auth/sign-in` and `profiles` were working (albeit an empty database for profiles).

madcampos: That is interesting... The mocks should have some way of putting the data the db will return, but again, I haven't tested it 🙈🙊

src/routes/auth/data.test.ts
```diff
@@ -0,0 +1,128 @@
...
+describe('Auth Data Functions', () => {
+	let env: MockEnvBindings;
+
+	beforeEach(() => {
+		env = new MockEnvBindings();
+	});
+
+	describe('getLoginInfo', () => {
+		test('returns login info for activated, non-deleted user', async () => {
+			env.setDatabase({
```
lylaminju, line 30: Questions 🤔 

1. Is it appropriate to use `setDatabase()` to set `first<T>():T` for the unit test ?
   - To implement isolated unit test for individual function, I guess setting `first()` is reasonable, but I'm not sure.

2. Is it the best practice to use generic type "T" when declaring `first()` ?
3. If it is the best practice,, is there any way to avoid using `/* eslint-disable @typescript-eslint/consistent-type-assertions */`? I had to add this to prevent eslint errors related to `as T`

src/routes/auth/data.test.ts
```diff
@@ -0,0 +1,128 @@
...
+				first: <T>() =>
+					({
+						id: sampleUser.id,
+						password: sampleUser.password,
+						access: sampleUser.access
+					}) as T
+			});
```
lylaminju: Questions 🤔 

1. Is it appropriate to use `setDatabase()` to set `first<T>():T` for the unit test ?
   - To implement isolated unit test for individual function, not interacting with real DB, I thought setting `first()` is reasonable, but I'm not sure.

2. Is it the best practice to use generic type "T" when declaring `first()` ?
3. If it is the best practice,, is there any way to avoid using `/* eslint-disable @typescript-eslint/consistent-type-assertions */`? I had to add this to prevent eslint errors related to `as T`

bcha92: 1. I don't have an answer for that but if this avoids running into errors with cloudflare workers (for some reason we still have issue with due to circumstances beyond our control), we'll consider it a temporary fix for now.
2. In practice, we really shouldn't be using generic types, but tesing some edge cases is not a bad idea at all.
3. We should find a way to avoid using `eslint-disable`. It's normally considered bad practice, but in certain circumstances where we are testing edge cases or strict assertion, this is unavoidable.

madcampos: > Is it appropriate to use setDatabase() to set first<T>():T for the unit test ?

If this comes from the mock I wrote, yes. We need to add a reset with a `beforeEach` call or something like that to reset between tests.

> Is it the best practice to use generic type "T" when declaring first() ?

No, this should be a reference to the actual type so the tests correspond to the expected type shape. If we only need a very partial thing we can write some type-foo and use `Partial`, `Omit`, `Pick`, and friends.

Worst case scenario we tell typescript to STFU with the comment `//@ts-expect-error`.

> If it is the best practice,, is there any way to avoid using /* eslint-disable @typescript-eslint/consistent-type-assertions */? I had to add this to prevent eslint errors related to as T

We should be using `assert` instead of `as` as much as possible. That being said, if eslint is getting on the way, telling it to STFU with an `eslint-ignore` comment is fine. Just don't overuse it and don't ignore everything 🙃

> I don't have an answer for that but if this avoids running into errors with cloudflare workers (for some reason we still have issue with due to circumstances beyond our control), we'll consider it a temporary fix for now.

We can give cloudflare's native testing thing a try again in the future. But the main issue is their testing tool doesn't create tables nor seed the database without lots of hacks.

I was planning on sending a bug report but got too lazy to write a minimal reproduction of the bug. 🙈

> In practice, we really shouldn't be using generic types, but tesing some edge cases is not a bad idea at all.

We shouldn't be casting to a _generic type_ here, the problem is casting to `as T`.

It would be better, if possible, to cast to the actual type and use `satisfies` instead of `as` for better catching things that the type cast may hide.

When we use `as` we literally say to typescript "I know what I'm doing, this is the type, don't question".

On the other hand, when we use `satisfies` we tell typescript "I know what to expect here, and it have this shape. Tell me if there is a mismatch though".

> We should find a way to avoid using eslint-disable. It's normally considered bad practice, but in certain circumstances where we are testing edge cases or strict assertion, this is unavoidable.

Yup, 100%

bcha92: resolving as it is outdated

madcampos (dismissed):
I've added cloudflare's vitest helper. That fixes the issue with vitest not finding the `cloudflare:workers` module.

Also fixed one of the tests and added a `skip` to the other for now.

